PR #56, opened 2025-04-15
Update size of the mark on the app icon to match sizing of other ios …

marksftw:
…apps


marksftw:
Old on left, New on right

IMG_4172

PR #99, opened 2025-06-08
feat: add system prompt indicator to chat interface

### LINKED ISSUE #98: Show System Prompt
If one of the user's chat had a system prompt, show a subtle indicator somewhere where they can click it to view what the system prompt was, with a button to copy the system prompt too maybe. 

I'm not sure if it should be at the top of the chat or in the chat title area, but use your best judgement and make it modern and nice looking, but keep in mind that it's not going to be a primary action. Just something we need for users to occasionally see and know what their system prompt was for that chat, so they can copy it to another chat later.

AnthonyRonning:
Add system prompt indicator feature to help users view and copy system prompts between chats.

## Changes
- Created SystemPromptIndicator component with subtle Bot icon and "System" label
- Shows dialog with full system prompt content when clicked
- Includes copy functionality with visual feedback
- Only appears when chat has a system prompt
- Integrated near chat title for easy discoverability

Resolves #98

### DIFF
--- a/frontend/src/components/SystemPromptIndicator.tsx
+++ b/frontend/src/components/SystemPromptIndicator.tsx
@@ -1,0 +1,90 @@
+import { useState, useCallback } from "react";
+import { Bot, Copy, Check } from "lucide-react";
+import {
+  Dialog,
+  DialogContent,
+  DialogDescription,
+  DialogHeader,
+  DialogTitle,
+  DialogTrigger
+} from "@/components/ui/dialog";
+import { Button } from "@/components/ui/button";
+import { cn } from "@/utils/utils";
+
+interface SystemPromptIndicatorProps {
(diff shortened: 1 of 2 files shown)


marksftw:
@claude on mobile the new button pushes the chat title off the screen to the left. See the screenshot. Put the new system button next to the “Person” icon in the prompt area instead. Directly to the right of the “Person” icon. 
IMG_5249

marksftw:
Nailed it. Lgtm. 
IMG_5260

AnthonyRonning:
@claude this doesn't look good on desktop. it's making the entire first chat aligned way too much to the right. is there a way to make this look better? 

Screenshot 2025-06-10 at 12 29 27 AM

Also it should only show the system message at the very top for the first message. existing user messages don't need to show it.

PR #123, opened 2025-06-16
fix: Add email verification check before Stripe checkout

### LINKED ISSUE #122: Bug: User paying without verifying their email first
A user paid without verifying their email first. This was probably due to the marketing and pricing pages for non-logged in users. if they click the purchase button while not logged in, they it'll prompt them to sign up. and then after signing up, it'll redirect them straight to the pricing page and I think even open stripe automatically. 

This works in like 99% of cases and even verifying email after sign up is _okay_ but we've had a case where we've had email delivery to a specific user after they paid. 

Can we do some sort of check on the pricing page, similar to the logged in home route for logged in users that check if they have a verified email before redirecting them to stripe if they're coming straight from the marketing/pricing page while being not logged in?

AnthonyRonning:
Fixes #122

Adds email verification check on the pricing page to prevent users from paying without verifying their email first.

## Changes
- Added email verification check in pricing page before Stripe redirect
- Shows error message if email not verified
- Added VerificationModal to pricing page for easy verification
- Prevents payment without email verification

## Summary by CodeRabbit

- **New Features**
  - Added an email verification check before allowing users to proceed with subscription checkout.
  - Introduced a notification prompting users to verify their email if it is not verified.

- **User Interface**
  - Integrated a verification modal on the pricing page to assist with email verification.

### DIFF
--- a/frontend/src/routes/pricing.tsx
+++ b/frontend/src/routes/pricing.tsx
@@ -13,6 +13,7 @@ import { Button } from "@/components/ui/button";
 import { Switch } from "@/components/ui/switch";
 import { type } from "@tauri-apps/plugin-os";
 import { PRICING_PLANS } from "@/config/pricingConfig";
+import { VerificationModal } from "@/components/VerificationModal";
 
 // File type constants for upload features
 const SUPPORTED_IMAGE_FORMATS = [".jpg", ".png", ".webp"];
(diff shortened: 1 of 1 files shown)


AnthonyRonning:
@claude - i didn't want just an email verification "check" - i also wanted it to do the same popup modal for verifying email address that we do on the logged in home route too. 

also what coderabbit is relevant too: "In frontend/src/routes/pricing.tsx around lines 332 to 339, the current logic
sets checkoutError and returns early when the user's email is unverified, which
causes the entire pricing UI to be replaced by an error screen and hides the
VerificationModal. To fix this, modify the code to avoid returning early after
setting checkoutError and ensure that the VerificationModal component is
rendered alongside the error message. Adjust the rendering logic so that when
checkoutError is set due to email verification, the VerificationModal is shown
to allow users to verify their email without needing a page reload." 

I didn't want it to show as an error

PR #151, opened 2025-07-15
fix: iOS external links now open properly using Tauri opener plugin

### LINKED ISSUE #149: Website links at the bottom of the iPhone app don't work, except for Downloads
In the iPhone app, when you tap the "Learn More", "OpenSecret", "Privacy Policy", and "Terms of Service" buttons at the bottom of the page, they don't do anything. If the user does a long press, then they load the link. The Downloads button does work correctly.

The broken buttons attempt to open the link in a new tab, but this is an iPhone app. It should open the link in the default web browser of the device.

We are using Tauri 2 for our iPhone app, so you might need to go through that to handle this, maybe even a plugin.

Keep the Downloads button and all other buttons as they are, don't change their behavior.

I circled the broken buttons in this screenshot.

Image

AnthonyRonning:
- Created openExternalLink utility function for cross-platform link handling
- Created ExternalLink React component for consistent link behavior
- Replaced all window.open calls with the new utility
- Updated all external anchor tags to use ExternalLink component
- Ensures iOS users can open links without long-pressing

Fixes #149

### DIFF
--- a/frontend/src/components/AccountMenu.tsx
+++ b/frontend/src/components/AccountMenu.tsx
@@ -10,6 +10,7 @@ import {
 } from "lucide-react";
 
 import { Button } from "@/components/ui/button";
+import { openExternalLink } from "@/utils/openExternalLink";
 import {
   DropdownMenu,
   DropdownMenuContent,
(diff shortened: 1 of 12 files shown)


AnthonyRonning:
@testflight build

PR #186, opened 2025-08-09
feat: implement default model selection based on user plan

### LINKED ISSUE #184: Default Model Selection
The default LLM selected for the user can vary depending on the scenario. Here are a few situations to handle:

1. New user on the Free plan. Their only model is Llama 3.3
2. New user on a Starter plan. They have access to Gemma 3. Select that one as default.
3. New user on Pro, Max, or Team. They have access to all of the models. Select GPT-OSS 120B as the default.
4. Existing user who has already chatted, select the last model used.
5. Existing user who wants to set their own default. They can set this to a specific model or "Last model used" which will be the selection when they first look at the selections. Currently Maple does not have a way for the user to pick a default model in their Profile, that would need to be augmented.

marksftw:
Closes #184

Implements default model selection based on user subscription plan:

- Free plan users default to Llama 3.3 70B
- Starter plan users default to Gemma 3 27B
- Pro/Max/Team plan users default to GPT-OSS 120B
- Existing users' last used model takes priority over plan defaults
- Model selection is automatically persisted to localStorage

## Summary by CodeRabbit

* **New Features**
  * Model selection is now saved and restored across sessions, so your last used model is automatically reselected when you return.
  * Model selection dynamically adapts to changes in your billing status.

* **Improvements**
  * Access to models is now managed more consistently based on your billing plan.
  * Default model selection is smarter, taking into account your previous choices and current billing plan.

### DIFF
--- a/frontend/src/components/ModelSelector.tsx
+++ b/frontend/src/components/ModelSelector.tsx
@@ -12,80 +12,10 @@ import { useEffect, useRef } from "react";
 import { useNavigate } from "@tanstack/react-router";
 import type { Model } from "openai/resources/models.js";
 import { hasAccessToModel as checkModelAccess } from "@/utils/modelDefaults";
+import { MODEL_CONFIG, DEFAULT_TOKEN_LIMIT, getModelTokenLimit } from "@/utils/modelConfig";
 
-// Model configuration for display names, badges, and token limits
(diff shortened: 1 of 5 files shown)


AnthonyRonning:
closing in favor of #190

PR #211, opened 2025-08-23
feat: hide encryption box after 3 chats and move verified button to footer

### LINKED ISSUE #210: Stop showing the "Encrypted. At Every Step" box after a while
Feedback from a user is that the "Encrypted. At every step" box becomes noise after a while. That space could be empty or eventually we could put helpful suggestions or other features there.

Idea:
- Move the "Verified" button out of the box and put it in the footer next to the "All services are online" badge. Use the same green checkmark that the "All services are online" badge is using. Basically style "Verified" exactly like that other badge and put them side-by-side. Have the "Verified" badge open the same page it does now `/proof`, but in a new tab
- Display the "Encrypted. At every step." box until they have made at least three chats. After that stop showing it.
- On desktop, when the "Encrypted." box is gone, maintain an empty space of the same size so the footer of the page stays down at the bottom. On mobile don't maintain an empty space so the footer is visible on the screen.

marksftw:
**Summary**
- Hide "Encrypted. At every step" box after user has made 3+ chats
- Move "Verified" button to footer next to "All services are online" badge
- Maintain desktop layout spacing, optimize mobile spacing

**Changes**
- Added chat counting mechanism via useChatCount hook
- Created ConditionalInfoCard for responsive box hiding
- Updated SimplifiedFooter to include VerificationStatus
- Clean component separation and proper responsive design

Resolves #210

## Summary by CodeRabbit

* **New Features**
  * Live verification status added to the footer with dynamic feedback and a proof link that opens in a new tab.
  * Conditional info card on the home page that hides after a few chats; preserves desktop spacing while removing extra space on mobile.

* **Improvements**
  * Explainer content simplified and centered the “Learn more” link.
  * Better responsive spacing and safer cancellation handling to prevent UI flashes.

### DIFF
--- a/frontend/src/components/ConditionalInfoCard.tsx
+++ b/frontend/src/components/ConditionalInfoCard.tsx
@@ -5,9 +5,19 @@ import { useChatCount } from "@/hooks/useChatCount";
 export function ConditionalInfoCard() {
   const { hasMinChats, isLoading } = useChatCount();
 
-  // While loading, don't render anything to prevent flash
+  // While loading, render transparent placeholder to maintain footer position
   if (isLoading) {
-    return null;
+    return (
+      <>
+        {/* Desktop: Transparent placeholder with same height */}
+        <div className="hidden md:block" aria-hidden="true">
+          <Card className="bg-transparent border-transparent shadow-none">
+            <CardHeader className="py-[88px]" />
+          </Card>
+        </div>
+        {/* Mobile: No placeholder needed */}
+      </>
+    );
   }
 
   return (


marksftw:
@claude on Desktop, it is still having the layout problem. I attached two screenshots. "first-load" shows how it looks when the page loads. The footer is up, close to the text box. Then it jumps down into place in "second-after-load". This is incorrect behavior. For Desktop the footer should already be in place on page load and not move after being rendered and shown to the user.

first-load

PR #216, opened 2025-08-26
fix: prevent layout shifts in code blocks during streaming

### LINKED ISSUE #209: Output gets jumpy
It seems that when output gets particularly long and includes different formatting, it gets jumpy. Once the LLM is done returning data, the flashing stops.

Browsers tested:
- Chrome
- Safari

Models tested:
- gpt-oss-120b
- DeepS
- eek R1 0528 671b

This prompt will cause the flashing behavior 

```
Tell me a story. 12 minute read. Insert two blocks of javascript code in the middle of the story. One that shows how to reverse the characters in a string and later one that shows how to store capitol cities for US states in a data structure. Put section headers on the story using markdown. Include a markdown table in the output.
```

Here is a video of the behavior: 

https://github.com/user-attachments/assets/527fe524-7068-43c6-bd10-c8cb509117bb

AnthonyRonning:
Fixed critical issue where code blocks would cause the entire layout to jump/flash during message streaming, particularly on mobile. This was causing accessibility issues.

The root cause was that as content streamed in, the browser would recalculate whether horizontal scrolling was needed, changing the code block height and causing layout reflow.

Changes:
- Set code elements to use width: max-content with min-width: 100% to maintain consistent dimensions
- Changed display from inline-block to block for proper width calculations
- Set overflow-x: visible on code elements to let parent handle scrolling
- Added proper width constraints to pre elements
- Wrapped PreCode component in full-width container
- Disabled transitions to prevent animation flicker

This ensures code blocks maintain consistent height regardless of content length, preventing the disorienting jumping effect during streaming.

Fixes #209 

## Summary by CodeRabbit

* **Style**
  * Improved Markdown code block layout: full-width display with horizontal scrolling only, no vertical scroll.
  * Ensured code content doesn’t clip and respects padding for consistent sizing.
  * Disabled transitions on highlighted code for snappier rendering.

* **Bug Fixes**
  * Prevented layout breakage by allowing long words/tokens to wrap in Markdown.
  * Removed unintended scrollbars and clipping inside code blocks for clearer readability.

### DIFF
--- a/frontend/src/chat.css
+++ b/frontend/src/chat.css
@@ -774,24 +774,30 @@
 .markdown-body .highlight pre,
 .markdown-body pre {
   padding: 16px 16px 8px 16px;
-  overflow: auto;
+  overflow-x: auto;
+  overflow-y: hidden;
   font-size: 85%;
   line-height: 1.45;
   border-radius: 6px;
   direction: ltr;
+  width: 100%;
+  min-width: 100%;
+  box-sizing: border-box;
 }
 
 .markdown-body pre code,
 .markdown-body pre tt {
-  display: inline-block;
-  max-width: 100%;
+  display: block;
   padding: 0;
   margin: 0;
-  overflow-x: scroll;
+  overflow-x: visible;
   line-height: inherit;
   word-wrap: normal;
   background-color: transparent;
   border: 0;
+  white-space: pre;
+  width: max-content;
+  min-width: 100%;
 }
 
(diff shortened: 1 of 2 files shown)


marksftw:
Works great in Safari on Mac, iOS, and the TestFlight build. 

Only small thing I noticed is that I can’t horizontally scroll in a code block until the full output is done. Understandable. I’m okay with that.

PR #271, opened 2025-10-13
Add DeepSeek V3.1 Terminus model to model selector

AnthonyRonning:
## Summary by CodeRabbit

* **New Features**
  * Added "DeepSeek V3.1 Terminus" to the model selector (Pro-only, 130k token limit; badges: Pro, New).
  * UI now renders a "Reasoning" badge with orange-gradient styling.

* **Updates**
  * Renamed "DeepSeek R1" display label, marked Pro-only with a 130k token limit and added the "Reasoning" badge.
  * Removed the "New" badge from two existing Pro models.

### DIFF
--- a/frontend/src/components/ModelSelector.tsx
+++ b/frontend/src/components/ModelSelector.tsx
@@ -43,9 +43,9 @@ export const MODEL_CONFIG: Record<string, ModelCfg> = {
     tokenLimit: 20000
   },
   "deepseek-r1-0528": {
-    displayName: "DeepSeek R1 0528 671B",
+    displayName: "DeepSeek R1 671B",
     shortName: "DeepSeek R1",
-    badges: ["Pro", "New"],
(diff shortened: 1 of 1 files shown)


AnthonyRonning:
@claude can we remove the "new" tag on all of the other models besides deepseek v3.1 ? maybe we can add a custom "Reasoning" tag on the other deepseek model. I think we use to allow for custom-like tags, i think it was orange or something.

AnthonyRonning:
@claude i accidentally force pushed and lost the last two changes you  made. can you try these again? 

"can we remove the "new" tag on all of the other models besides deepseek v3.1 ? maybe we can add a custom "Reasoning" tag on the other deepseek model. I think we use to allow for custom-like tags, i think it was orange or something."

and 

"let's also drop the 0528 part from that model name in the dropdown selector, it's making the text too much and its dropping down to a new line"

PR #367, opened 2026-01-06
feat: add pull-to-refresh to chat history sidebar

### LINKED ISSUE #366: Pull down to refresh
Please implement a "pull down to refresh" logic in the history list so that I can instantly refresh the conversations list api call. we do not need to implement this for the old archived chat type.

AnthonyRonning:
Implements pull-to-refresh functionality for the chat history list in the sidebar.

## Changes
- Add pull-to-refresh for conversations list
- Works on touch devices (mobile) and mouse (desktop)
- Only activates when scrolled to the top of sidebar
- Minimalistic refresh icon with smooth animations
- Uses existing pollForUpdates() function for instant refresh
- Pull threshold: 60px with resistance factor
- Cross-platform: iOS, Android, desktop, and web
- No pull-to-refresh on archived chats (as requested)

Fixes #366

## Summary by CodeRabbit

* **New Features**
  * Adds pull-to-refresh to the chat history list with touch, mouse drag, and wheel support, platform-aware input handling, resistance threshold, and an animated refresh indicator during pull and refresh.

* **API Changes**
  * The chat history component now accepts an optional container reference prop to enable the pull-to-refresh integration.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/frontend/src/components/ChatHistoryList.tsx
+++ b/frontend/src/components/ChatHistoryList.tsx
@@ -91,6 +91,7 @@ export function ChatHistoryList({
   const pullDistanceRef = useRef(0);
   const wheelDeltaAccumulator = useRef(0);
   const lastRefreshTime = useRef(0);
+  const isRefreshingRef = useRef(false);
 
   // Fetch initial conversations from API using the OpenSecret SDK
   const { isPending, error } = useQuery({
@@ -185,6 +186,7 @@ export function ChatHistoryList({
 
   // Pull-to-refresh handler
   const handleRefresh = useCallback(async () => {
+    isRefreshingRef.current = true;
     setIsPullRefreshing(true);
     try {
       await pollForUpdates();
@@ -195,6 +197,7 @@ export function ChatHistoryList({
       setTimeout(() => {
         setIsPullRefreshing(false);
         setPullDistance(0);
+        isRefreshingRef.current = false;
         // Record the time when refresh completes for cooldown
         lastRefreshTime.current = Date.now();
       }, 300);
@@ -211,14 +214,14 @@ export function ChatHistoryList({
 
     const handleTouchStart = (e: TouchEvent) => {
       // Only start pull if we're at the top of the scroll
-      if (container.scrollTop === 0 && !isPullRefreshing) {
+      if (container.scrollTop === 0 && !isRefreshingRef.current) {
         pullStartY.current = e.touches[0].clientY;
         isPulling.current = true;
       }
     };
 
     const handleTouchMove = (e: TouchEvent) => {
-      if (!isPulling.current || isPullRefreshing) return;
+      if (!isPulling.current || isRefreshingRef.current) return;
 
       const currentY = e.touches[0].clientY;
       const distance = currentY - pullStartY.current;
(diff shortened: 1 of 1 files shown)


AnthonyRonning:
@claude the code review agents (greptile and coderabbit) had some comments. can you handle the ones that are important if they are easy and need to be fixed?

Also this works well on web if I hold down the mouse and pull down. However it should just work if I try "scrolling up" when I'm already on the top.

So maybe you can have the logic so that it is a "press and pull down" if it's in the mobile view, but if it's in the full screen desktop view, it's simply a "scroll up to refresh" ? if that makes sense

AnthonyRonning:
@claude - two things. 

This is a comment left during a code review.
Path: frontend/src/components/ChatHistoryList.tsx
Line: 307:315

Comment:
**style:** The conditional listener registration adds either wheel events (desktop) or mouse events (mobile), but cleanup always removes both sets. While this works, it wastes cycles removing listeners that were never added. Consider matching cleanup logic to registration:

```
return () => {
  container.removeEventListener("touchstart", handleTouchStart);
  container.removeEventListener("touchmove", handleTouchMove);
  container.removeEventListener("touchend", handleTouchEnd);
  
  if (isDesktopPlatform) {
    container.removeEventListener("wheel", handleWheel);
  } else {
    container.removeEventListener("mousedown", handleMouseDown);
    window.removeEventListener("mousemove", handleMouseMove);
    window.removeEventListener("mouseup", handleMouseUp);
  }
};
```

<sub>Note: If this suggestion doesn't match your team's coding style, reply to this and let me know. I'll remember it for next time!</sub>

How can I resolve this? If you propose a fix, please make it concise.

---

for two, on desktop view it doesn't work at all. it seems to want to pull down the whole app view when I try to scroll up. Not just restricted to the sidebar when my mouse is hovered over the sidebar. Because of this, it doesn't work at all with the full desktop / web view.

AnthonyRonning:
@claude please look at the comment by greptile: 

```
This is a comment left during a code review.
Path: frontend/src/components/ChatHistoryList.tsx
Line: 296:302

Comment:
**style:** On desktop, any upward scroll attempt at the top triggers refresh immediately. This may be too sensitive—users scrolling up quickly might accidentally trigger refreshes. Consider requiring a longer scroll distance or multiple scroll events before triggering.

```suggestion
    // Desktop: detect scroll up when already at top (overscroll)
    let wheelDeltaAccumulator = 0;
    const WHEEL_THRESHOLD = -50; // Require some accumulated scroll
    
    const handleWheel = (e: WheelEvent) => {
      if (!isDesktopPlatform || isPullRefreshing) return;

      // Only handle if the event target is within our container
      if (!container.contains(e.target as Node)) return;

      // Check if we're at the top and trying to scroll up
      if (container.scrollTop === 0 && e.deltaY < 0) {
        // Prevent default to avoid browser overscroll bounce
        e.preventDefault();
        
        wheelDeltaAccumulator += e.deltaY;
        if (wheelDeltaAccumulator <= WHEEL_THRESHOLD) {
          wheelDeltaAccumulator = 0;
          handleRefresh();
        }
      } else {
        wheelDeltaAccumulator = 0;
      }
    };
```

<sub>Note: If this suggestion doesn't match your team's coding style, reply to this and let me know. I'll remember it for next time!</sub>

How can I resolve this? If you propose a fix, please make it concise.
```


also, yes it works but i think my confusion is that there's no nice animation like we had for mobile :( i liked that refresh indicator

AnthonyRonning:
@claude i think we're getting close! one minor thing. idk how easy this will be to fix. but if I do a "long swipe" while on the web, it does the refresh properly but it keeps doing it for a couple iterations. I think because technically, the scroll is **still** happening past the refresh, so it does it 2 or 3 more times immediately in a row. I'm not sure how easy it would be to detect this and fix it so that even if the user does one long "scroll" it is only accounted for once

AnthonyRonning:
@claude that didn't seem to make a difference. it refreshes right after it finishes since it's still technically scrolling. what if we undid that last change and just "lock" it from refreshing for 5 seconds after it finishes a refresh? that would prevent a long scroll from triggering it again right?

AnthonyRonning:
@claude nope it is severely broken now. very buggy and screwed up.

PR #420, opened 2026-02-09
feat: Add approved version check for API Management button on iOS/Android

### LINKED ISSUE #419: Explose the api management tab on ios/mobile
In billing / pricing routes/pages we check to see if the current version of the app is an "approved version" before allowing purchasing. 

We should do the same check in the sidebar's account "API Management" button. We should enable this if maple is on or before an approved version, following the same logic as the pricing routes.

AnthonyRonning:
## Summary

This PR implements the approved version check for the API Management button in the sidebar, following the same logic used for the pricing routes.

## Changes

- Modified `frontend/src/components/AccountMenu.tsx` to add version checking logic
- API Management button now shows on desktop/web (always)
- On iOS/Android, it only shows if the app version is approved by the server
- Fetches products with version to check `is_available` status

## Testing

- ✓ Prettier formatting passed
- ✓ ESLint linting passed
- ✓ TypeScript build passed
- ✓ All pre-commit hooks passed

Resolves #419

---

<a href="https://app.devin.ai/review/opensecretcloud/maple/pull/420" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open with Devin
  </picture>
</a>

## Summary by CodeRabbit

* **Improvements**
  * API Management menu is now platform-aware: on desktop it remains visible consistently; on mobile it appears only when relevant product(s) are available, preventing empty or inapplicable menu items.
  * Enhances reliability of menu visibility across devices for a clearer navigation experience.

### DIFF
--- a/frontend/src/components/AccountMenu.tsx
+++ b/frontend/src/components/AccountMenu.tsx
@@ -13,7 +13,7 @@ import {
   FileText,
   ChevronLeft
 } from "lucide-react";
-import { isMobile, isTauri, isIOS, isAndroid } from "@/utils/platform";
+import { isMobile, isTauri, isIOS } from "@/utils/platform";
 
 import { Button } from "@/components/ui/button";
 import {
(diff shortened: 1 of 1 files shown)


AnthonyRonning:
@claude is this comment valid or not? we should have our existing ways that we handle useQuery and should have been using it for pricing related api calls throughout the pricing page. I care more about our existing standards and correctness and whether or not our approach works, vs whatever this bot is saying. research this in our codebase. ignore with an explanation if it's invalid. or make the fix if it is valid:

```

🟡 Shared query key with error-swallowing queryFn can poison the pricing page's product cache

The new useQuery in AccountMenu uses query key ["products", isIOSPlatform, isAndroidPlatform] with a queryFn that catches errors and returns null as successful data. The pricing page at frontend/src/routes/pricing.tsx:267 uses the exact same query key but its queryFn properly throws errors so React Query can show an error state.

Detailed Explanation of Cache Poisoning
On mobile, if AccountMenu's product fetch fails (e.g., network issue), the catch block at AccountMenu.tsx:147-149 returns null, which React Query caches as successful data under the shared key ["products", true, false] (iOS) or ["products", false, true] (Android).

When the user then navigates to the pricing page, React Query serves the cached null immediately as stale data. The pricing page receives products = null, productsError = undefined, and productsLoading = false. This bypasses both the loading skeleton (pricing.tsx:649) and the error UI (pricing.tsx:682), resulting in an empty pricing page with no products and no error message shown to the user.

While React Query's default staleTime: 0 means a background refetch will start, during the window before it completes the user sees a broken UI. If the refetch also fails, the pricing page's queryFn would properly set the error state, but the initial flash of empty content is still a poor user experience.

Fix: Use a distinct query key (e.g., ["products-version-check", ...]) for the AccountMenu query, or avoid swallowing errors and instead let React Query handle the error state naturally.
```

PR #421, opened 2026-02-11
Projects Prototype (Frontend-only, localStorage)

marksftw:
## Projects Prototype (Frontend-only, localStorage)

Video Walkthrough: 

https://github.com/user-attachments/assets/540e9bae-aacb-4c45-9867-85ae6f049a92

  Adds a complete Projects feature prototype to evaluate the UX before building
   the backend. All data is persisted in localStorage — no API changes.

  ### What it does

  - **Organize chats into projects** — create, rename, and delete projects from
   the sidebar or project detail page
  - **Custom instructions per project** — set context that applies to all chats
   within a project (chats don't share history with each other, but they all
  use the project's custom instructions and files as context)
  - **File attachments per project** — add reference files to a project that
  are available across all its chats
  - **Unified chat context menu** — rename, delete, move between projects, or
  remove from a project, consistent everywhere (sidebar + project page)
  - **Collapsible sidebar sections** — both Projects and Recents are
  independently collapsible with persisted state
  - **Start chats from project page** — new chats are automatically assigned to
   the project

  ### How it works

  Projects are stored in localStorage (`maple_projects`). When a chat is
  started from a project page, a `project_id` URL param triggers
  auto-assignment after conversation creation via a custom event. Chats
  assigned to projects are filtered out of the Recents list. Deleting a project
   marks its chats as deleted.

  ### New files (10)
  - `state/ProjectsContext.tsx` + `state/useProjects.ts` — state management
  - `components/ChatContextMenu.tsx` — shared context menu used across sidebar
  and project page
  - `components/projects/` — ProjectsList, ProjectDetailPage, and 4 dialog
  components
  - `routes/_auth.project.$projectId.tsx` — project detail route

  ### Modified files (5)
  - `app.tsx` — added ProjectsProvider
  - `Sidebar.tsx` — projects section, create-project-for-chat event handling
  - `ChatHistoryList.tsx` — collapsible Recents, exclude project chats, shared
  context menu
  - `UnifiedChat.tsx` — auto-assign chat to project via URL param

---

<a href="https://app.devin.ai/review/opensecretcloud/maple/pull/421" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open with Devin
  </picture>
</a>

## Release Notes

* **New Features**
  * Added project management system to organize and group chats
  * Create, rename, and delete projects with dedicated detail pages
  * Move chats between projects or remove them from projects
  * Set custom instructions per project
  * Attach and manage files within projects
  * Access chat context menus with project-related actions

### DIFF
--- a/frontend/src/app.tsx
+++ b/frontend/src/app.tsx
@@ -7,6 +7,7 @@ import { routeTree } from "./routeTree.gen";
 import { useOpenSecret, OpenSecretProvider } from "@opensecret/react";
 import { OpenAIProvider } from "./ai/OpenAIContext";
 import { LocalStateProvider } from "./state/LocalStateContext";
+import { ProjectsProvider } from "./state/ProjectsContext";
 import { ErrorFallback } from "./components/ErrorFallback";
 import { NotFoundFallback } from "./components/NotFoundFallback";
 import { BillingServiceProvider } from "./components/BillingServiceProvider";
@@ -96,6 +97,7 @@ export default function App() {
         }}
       >
         <LocalStateProvider>
+          <ProjectsProvider>
           <OpenAIProvider>
             <QueryClientProvider client={queryClient}>
               <TooltipProvider>
@@ -110,6 +112,7 @@ export default function App() {
               </TooltipProvider>
             </QueryClientProvider>
           </OpenAIProvider>
+          </ProjectsProvider>
         </LocalStateProvider>
       </OpenSecretProvider>
     </NotificationProvider>
--- a/frontend/src/components/ChatContextMenu.tsx
+++ b/frontend/src/components/ChatContextMenu.tsx
@@ -22,10 +22,11 @@ import {
 interface ChatContextMenuProps {
   chatId: string;
   isMobile: boolean;
-  // Available projects for "Move to project" submenu
+  // All projects for "Move to project" submenu
   projects: { id: string; name: string }[];
-  // If set, shows "Remove from {projectName}"
+  // If set, shows "Remove from {projectName}" and grays out current project in submenu
   currentProjectName?: string;
+  currentProjectId?: string;
   // Optional callbacks — item only rendered when provided
   onSelect?: () => void;
   onRename?: () => void;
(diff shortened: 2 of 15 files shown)


marksftw:
> not bad, appreciate you doing that.
> 
> i have a few thoughts:
> 
> 1. shared files context, this significantly increases scope here. We really only support a single doc at a time. And it's part of the context. We're really not well equiped to be able to handle multi document, let along shared across chats.

For the UX perspective, it would be awesome to add multiple files into the context. From the technical perspective, I can understand the limitation right now. The feature is still very useful without the file attachment feature. Would we be able to do a single file or should we just leave it to the user to attach the file to a single chat like they already do?

> 2. it's weird that a chat list is showing up under the chatbox. i think the chatbox is better off at the bottom where it typically is and the chat list either above it or to the right panel too

Yeah I can see that. Both ChatGPT and Claude have it above like this, so I was following their lead. That said, putting the chat box below could be fine.


> 3. Mobile view. Has all of the flows and design been tested for a mobile size view?

Yes, I tested it on mobile and just pushed a fix for a small layout issue. Everything else works great on mobile.

AnthonyRonning:
I'm not a fan of attempting to extract just the chatbox out of the main display. This isn't an extractable component and it's designed this way for a reason. 

I think I would prefer there to be something like a folder icon on the chatbox that you can auto assign a new chat to a project. And maybe you can have a "new chat" button on the project page that jumps to the main page but preloads the correct project.

marksftw:
> I'm not a fan of attempting to extract just the chatbox out of the main display. This isn't an extractable component and it's designed this way for a reason.
> 
> I think I would prefer there to be something like a folder icon on the chatbox that you can auto assign a new chat to a project. And maybe you can have a "new chat" button on the project page that jumps to the main page but preloads the correct project.

I'm good with that solution. I'll adjust the prototype with that behavior.

AnthonyRonning:
hmm, looks like cloudflare didn't publish it? can you do an empty commit and push when you have a chance

PR #517, opened 2026-05-15
Windows port PR 1: compile + launch + unsigned CI

ldstreet:
## Summary

First of several PRs porting Maple to Windows. Lands the minimum needed to compile and launch on a clean Windows host plus a `windows-latest` CI job that produces an unsigned NSIS installer on every PR.

## Why these changes

**`tauri.conf.json` — explicit `bundle.targets`, NSIS per-user, WebView2 bootstrapper.** Tauri's WiX/msi template hard-codes HKLM + per-machine install, which would shadow the HKCU keys we rely on for `cloud.opensecret.maple://` deep-link handlers. Switching `bundle.targets` from `"all"` to an explicit list excludes msi from future Windows builds without affecting mac/linux output. `installMode: "currentUser"` keeps the installer UAC-free and consistent with HKCU registration. `webviewInstallMode: "downloadBootstrapper"` pins behavior so a future Tauri default change doesn't surprise users.

**`capabilities/default.json` — drop dead `$HOME/.config/maple/**` entries.** Those scope tokens don't resolve to anything meaningful on Windows and aren't used anywhere today — the JS never invokes the Tauri fs plugin and `proxy.rs` writes via raw `std::fs`, which bypasses capabilities entirely. Removing them everywhere is cleaner than gating per-OS.

**`proxy.rs` — Windows path arm via Tauri's path resolver.** The old code hardcoded `$USERPROFILE/.config/maple/proxy_config.json` on Windows, which is the wrong path convention. Plumbing `&AppHandle` through the helpers lets us use `app_config_dir()` for the canonical `%APPDATA%\cloud.opensecret.maple\`. macOS/Linux behavior is byte-identical — atomic migration + keyring storage is deliberately deferred to keep this PR small.

**`desktop-build.yml` — `windows-latest` job.** Mirrors the existing Linux job: pre-fetches ONNX Runtime so the ort crate skips its own download, runs `tauri-action` with `--bundles nsis`, uploads the `.exe`. Authenticode signing is deliberately out of scope here — adding it would block on certificate provisioning. Updater minisigning uses the existing repo secret so the build doesn't error out after producing the installer.

## Manual smoke

Built and launched natively on a Parallels Win11 ARM64 VM (not the CI x64 artifact, but proves the same milestone). Installer ran per-user, no UAC. App launched, login screen rendered.

One gotcha worth flagging for future contributors: a missing `frontend/.env.local` produces a silent white-screen because Vite bakes env vars into the bundle at build time. Captured for a follow-up.

## Out of scope (later PRs)

Code signing, proxy keyring + atomic config migration, TTS DLL bundling, deep-link install-mode hardening, UI/downloads copy, `release.yml` + auto-updater Windows entry, full acceptance test pass.

## Test plan

- [ ] `build-windows` job uploads NSIS artifact
- [ ] mac/linux jobs still pass (no regression from `proxy.rs` `AppHandle` plumbing or `bundle.targets` change)
- [ ] Artifact installs cleanly on Windows and launches to login

### DIFF
--- a/.github/workflows/desktop-build.yml
+++ b/.github/workflows/desktop-build.yml
@@ -224,3 +224,79 @@ jobs:
             frontend/src-tauri/target/release/bundle/appimage/*.AppImage
             frontend/src-tauri/target/release/bundle/deb/*.deb
           retention-days: 5
+
+  build-windows:
+    runs-on: windows-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Setup Bun
+        uses: oven-sh/setup-bun@v1
+        with:
+          bun-version: 1.2.2
+
+      - name: Install Rust
+        uses: dtolnay/rust-toolchain@stable
+
+      - name: Install sccache
+        run: |
+          choco install sccache --no-progress -y
+          sccache --version
+
+      - name: Cache sccache
+        uses: actions/cache@v4
+        with:
+          path: ~\AppData\Local\Mozilla\sccache
+          key: ${{ runner.os }}-sccache-windows-${{ hashFiles('**/Cargo.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-sccache-windows-
+            ${{ runner.os }}-sccache-
+
+      - name: Provide ONNX Runtime (Windows)
+        shell: bash
+        run: |
+          ORT_VERSION=1.22.0
+          ORT_ROOT="$RUNNER_TEMP/onnxruntime"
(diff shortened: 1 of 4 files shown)


ldstreet:
@AnthonyRonning Looks like I'll need workflow approval here. Keeping as draft for now since I'm expecting some iteration getting the CI correct

ldstreet:
Superseded by #520 (same branch pushed to upstream now that I have org write access — gets us proper CI with secrets available).

PR #536, opened 2026-05-22
Windows port PR 2: local dev experience

ldstreet:
## Summary

- **`setup-windows.ps1`** (2a) — idempotent bootstrap: VS Build Tools components, Rust 1.95.0 (CI pin), Node LTS (bun has no Win-ARM binary), Git for Windows (provides Git Bash for the ORT helper), `just` (justfile runner), VC++ redists, LLVM, and a `frontend/.env.local` template. Defensively strips a UTF-8 BOM from an existing `.env.local` (PS 5.1 default-BOM I/O is a notorious footgun).
- **`tauri-windows.ps1`** + **`tauri.windows.conf.json`** (2b) — one-command `just windows-build` / `just windows-dev` wrappers. Handles vcvarsall.bat lookup, `CARGO_TARGET_DIR` off-source (Parallels-safe), ORT helper invocation, and the bun→npm overlay. Wraps both native and cross-arch builds (`arch=arm64` / `x64` / `arm64_amd64`).
- **docs** (2c) — `docs/windows-build.md` end-to-end guide, `frontend/.env.local.template`, README (Prerequisites + Development sections mirroring macOS/Linux/iOS), CLAUDE.md (`.env.local` requirement note).

## Stacked on #520

Base branch is `windows/pr1-compile-launch`, not master. Wait for #520 to merge before this can land on master.

## Patterns mirrored from #520 / master

- **Rust 1.95.0 pin** via `-RustToolchain` parameter (matches `dtolnay/rust-toolchain` in CI)
- **ORT helper extended for arm64**: `onnxruntime-pins.sh` gets `windows_arm64_*` entries; `provide-windows-onnxruntime.sh` reads `ORT_TARGET_ARCH` (default `x64`, so CI behavior unchanged). `tauri-windows.ps1` derives it from `-Arch`.
- **iOS recipe-delegates-to-script idiom** (`just ios-build-onnxruntime` → script; same shape for `just windows-build` → `tauri-windows.ps1`)
- **README per-platform Prerequisites + Development sections** (added Windows blocks parallel to macOS/Linux/iOS)
- **Inline `# why:` comments on every non-obvious workaround** (MSYS2 path mangling, `--passive` vs `--quiet`, BOM-strip, etc.)

## Manual smoke

Validated end-to-end on Parallels Win11 ARM64: `git checkout` → `setup-windows.ps1` → `just windows-dev` → Maple launches to login screen with Vite hot-reload. `onnxruntime.dll` loads at runtime on native ARM64 (extends #520's x64-under-emulation data point).

## Out of scope

- Authenticode signing — PR 7 (MPLR-goufxxvn)
- Native ARM64 shipping decision — MPLR-xhfehqft spike, still todo (partial data now: build + app launch confirmed on ARM64; CI cost / WebView2 / ort native-arm64 coverage TBD)

## Test plan

- [ ] `build-windows` jobs (both PR + master push) pass — should be unaffected since `ORT_TARGET_ARCH` defaults to x64
- [ ] Fresh Win11 VM: `git clone` → `setup-windows.ps1` → `just windows-dev` succeeds in ~30 min (the \"Done when\" criterion for MPLR-rqofgxpz)
- [ ] `just windows-build` on ARM64 produces a working native NSIS installer

### DIFF
--- a/frontend/.env.local.template
+++ b/frontend/.env.local.template
@@ -1,0 +1,18 @@
+# Copy this file to frontend/.env.local before running `just dev`, `just
+# desktop-build`, `just windows-dev`, etc. The values below point at the
+# production enclave -- override the URLs for a non-prod backend.
+#
+# Vite bakes these values into the bundle at BUILD time, not runtime --
+# rerun the dev server or rebuild after editing. A missing or empty
+# VITE_OPEN_SECRET_API_URL produces a silent white-screen on launch.
+
+VITE_OPEN_SECRET_API_URL=https://enclave.trymaple.ai
+
+# Public OpenSecret project id. Optional; the app uses this value by default.
+VITE_CLIENT_ID=ba5a14b5-d915-47b1-b7b1-afda52bc5fc6
+
+# Override if pointing at a non-prod billing backend.
+# VITE_MAPLE_BILLING_API_URL=https://billing.opensecret.cloud
+
+# Force a specific model for dev. Leave commented to use the app's default.
+# VITE_DEV_MODEL_OVERRIDE=gpt-4o
--- a/scripts/tauri-windows.ps1
+++ b/scripts/tauri-windows.ps1
@@ -1,0 +1,179 @@
+#Requires -Version 5.1
+<#
+.SYNOPSIS
+  One-command wrapper for `tauri build` / `tauri dev` on Windows.
+
+.DESCRIPTION
+  Handles the vcvarsall.bat + ORT env-var dance so the developer doesn't
+  need a Developer PowerShell open, doesn't need to remember the ORT helper,
+  and doesn't blow up on cargo target dir when the repo lives on a Parallels
+  shared folder.
+
+  Steps:
+    1. Locate VS Build Tools install via vswhere.
+    2. Point CARGO_TARGET_DIR at %USERPROFILE%\maple-cargo-target (overridable
+       via -CargoTargetDir or $env:CARGO_TARGET_DIR) -- Parallels shared-folder
+       writes don't play with cargo's default target directory.
+    3. Source frontend/src-tauri/scripts/provide-windows-onnxruntime.sh via Git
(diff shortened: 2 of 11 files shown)


frontend/.env.local.template
AnthonyRonning, line 1: Unnecessary. https://github.com/OpenSecretCloud/Maple/blob/master/frontend/.env.example

scripts/tauri-windows.ps1
AnthonyRonning, line 1: I don't like this at all. Please use nix on windows.

AnthonyRonning (changes requested):
I'm not sure I like many of these changes, nix should be used instead.

scripts/tauri-windows.ps1
ldstreet: I'd definitely love to use nix to stay consistent with the other platforms but my understanding is windows doesn't support nix unless you run it in a linux subsystem but then you wouldn't be able to compile the windows artifacts you need. did you have a specific strategy in mind for using nix on windows?

frontend/.env.local.template
ldstreet: this points to a local backend which I don't expect us to really be running on a windows vm? to validate and test the windows build I think it probably makes sense to point to a prod/staging environment. but this file definitely feels like a duplicate, what if we point to `/.env.example` but in the windows setup we overwrite the value to be a real env?

ldstreet:
Closing this. Per the discussion on #520, the Windows dev-experience and CI/setup tooling here (`setup-windows.ps1`, `tauri-windows.ps1`, `tauri.windows.conf.json`, `docs/windows-build.md`, `.env.local.template`) is being redone as part of the new nixos reproducible build setup, so there's no point landing it separately.

Remaining Windows work is consolidating into a single feature-complete PR on #520 (app/runtime code only — proxy paths, keyring, deep links, UI copy, TTS).

PR #637, opened 2026-07-16
Add manual Agent project ordering

### LINKED ISSUE #633: Agent Mode: Add manual project ordering
## Summary

Stop reordering Agent Mode projects by recent activity. Keep the project list stable while the user works and let the user control its order by dragging project rows.

## Desired behavior

- Clicking a project or one of its task sessions does not change the project's position.
- The entire project row can be dragged to a new position without breaking its existing click, expand/collapse, or new-session actions.
- During a drag, show a temporary insertion line at the exact position where the project will land.
- On drop, apply the new order and reuse the existing project-list shuffle animation as the rows settle.
- A project first added to Agent Mode appears at the top without changing the relative order of existing projects.
- The user's project order persists locally for that account and device across app launches.
- Task sessions within each project continue to use the app's existing automatic ordering.

## Acceptance criteria

- [ ] Selecting a project or task session never moves its project in the list.
- [ ] A project row can be dragged above or below any other project.
- [ ] The insertion line updates during the drag and accurately previews the drop position.
- [ ] Dropping a project persists the new order and uses the existing shuffle animation.
- [ ] Canceling a drag or dropping outside a valid target leaves the saved order unchanged.
- [ ] Existing row actions remain usable and do not accidentally trigger a reorder.
- [ ] New projects appear at the top while existing projects retain their relative order.
- [ ] The saved order is restored after relaunching Maple on the same device.
- [ ] Task-session ordering remains unchanged.
- [ ] Existing reduced-motion behavior is preserved.

## Reference screenshots

Four screenshots are attached to illustrate the expected resting order, new-project placement, draggable row, insertion marker, and drop animation.

1. Image
2. Image
3. Image
4. Image

## Implementation note

Project recency is currently applied both when recent roots are saved and when sidebar rows are sorted by last-used activity. Manual order should become the source of truth for projects, while the existing automatic task-session sort remains intact.

## Out of Scope

- Manual ordering of Projects in Chat Mode. This can be considered as a future extension after Agent Mode.
- Changes to the automatic ordering of task sessions within Agent projects.
- Cross-device synchronization of Agent projects, manual project order, or task-session history. These remain local to the current device.

marksftw:
https://github.com/user-attachments/assets/b21a44b7-bd15-44bd-b90a-0f8bc03bd3b8

Closes #633

## What

- Make Agent Mode project order user-controlled instead of activity-based.
- Add whole-project drag-and-drop with an insertion marker, click/drag threshold, cancellation, edge scrolling, and the existing settling animation.
- Persist order locally per account through a new Tauri command.
- Optimistically update the sidebar, rolling back and surfacing an error if persistence fails.
- Preserve automatic newest-first ordering for sessions within each project.

## Why

Project order was previously changed in two places: native recent-root persistence promoted projects during runtime and session activity, while the sidebar sorted projects by `lastUsedMs`. Selecting a project or session could therefore move its project unexpectedly.

## Impact

- Selecting, starting, or opening Agent sessions no longer moves projects.
- Newly added projects appear at the top without disturbing the existing order.
- Manual order survives relaunches for the same account and device.
- Persisted paths are validated and deduplicated, and stale requests cannot erase known roots.
- Chat Mode and task-session ordering remain unchanged.
- Reduced-motion behavior, row actions, disabled states, indicators, compact layouts, and scrolling remain supported.

## Checks

- `just format`
- `just lint` — 0 errors; 12 pre-existing warnings
- `just build`
- `cd frontend && bun test` — 119 passed, 0 failed
- `just rust-lint` — passed with existing Tao dependency warnings
- `cd frontend/src-tauri && cargo test --all-targets --locked` — 123 passed, 0 failed
- `git diff --check`
- Local debug Tauri app built and launched successfully

## Summary by CodeRabbit

* **New Features**
  * Added drag-and-drop reordering for project roots in the Agent sidebar, with live insertion/ghost feedback.
  * Persist project-root ordering and restore it across sessions and restarts.
  * Updated session grouping to reflect the selected project-root order and sort sessions newest-first.
* **Bug Fixes**
  * Prevent starting/restarting runtimes and creating sessions from reordering or mutating saved project-root history.
  * Improved resilience to legacy, stale, or corrupted stored project-root data.
* **Tests**
  * Added coverage for queued operation blocking and project-ordering merge/reorder/persistence behaviors.

### DIFF
--- a/frontend/src-tauri/src/agent.rs
+++ b/frontend/src-tauri/src/agent.rs
@@ -196,7 +196,7 @@ pub struct AgentRuntimeStatus {
     pub active_runs: HashMap<String, String>,
 }
 
-#[derive(Debug, Clone, Serialize, Deserialize)]
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
 pub struct RecentProjectRoot {
     pub path: String,
@@ -746,7 +746,9 @@ async fn start_runtime_for_user(
         *guard = Some(runtime);
     }
 
-    let _ = save_recent_project_root_inner(&app_handle, &user_id, &project_root);
+    // Starting a runtime is project use, not an explicit folder add. In particular, a
+    // session-derived root may be absent from a legacy capped recent-roots file; registering it
+    // here would incorrectly move that visible project to the top of the manual order.
     agent_config.default_project_root = Some(path_string(&project_root));
     agent_config.default_model = model;
     let _ = save_agent_config_inner(&app_handle, &user_id, &agent_config);
@@ -940,7 +942,22 @@ pub async fn agent_save_recent_project_root(
     let _runtime_lifecycle_guard = state.runtime_lifecycle.lock().await;
     ensure_account_generation(&state, &account_scope, generation).await?;
     let project_root = normalize_project_root(Path::new(&path))?;
-    save_recent_project_root_inner(&app_handle, &user_id, &project_root).map_err(|e| e.to_string())
+    register_explicit_project_root_inner(&app_handle, &user_id, &project_root)
+        .map_err(|e| e.to_string())
+}
+
+#[tauri::command]
+pub async fn agent_save_project_root_order(
+    app_handle: AppHandle,
+    state: State<'_, AgentRuntimeState>,
+    user_id: String,
+    paths: Vec<String>,
+) -> Result<Vec<RecentProjectRoot>, String> {
+    let account_scope = account_scope(&user_id)?;
+    let generation = account_generation(&state, &account_scope).await;
+    let _runtime_lifecycle_guard = state.runtime_lifecycle.lock().await;
+    ensure_account_generation(&state, &account_scope, generation).await?;
+    save_project_root_order_inner(&app_handle, &user_id, paths).map_err(|e| e.to_string())
 }
 
 #[tauri::command]
@@ -1039,7 +1056,8 @@ pub async fn agent_create_session(
         }
     }
     let summary = session_summary(&session);
-    let _ = save_recent_project_root_inner(&app_handle, &user_id, &root);
+    // Session creation must not mutate project order. Only explicit folder-add and reorder
+    // commands may change the persisted project list.
     let detail = AgentSessionDetail {
         session: summary.clone(),
         timeline: Vec::new(),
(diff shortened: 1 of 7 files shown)


marksftw:
## Future Projects UX opportunities

These are follow-up opportunities rather than blockers for this Agent Mode change.

### 1. Drag-ordering performance

The current implementation is deliberately correctness-first and should be fast for ordinary project lists. During a drag, however, each native pointer update measures every project header and updates React state, which rerenders the project/session subtree. If profiling shows jank as lists grow, the highest-value improvements would be:

- coalesce pointer movement through one `requestAnimationFrame` update;
- cache header/group geometry at drag start, refreshing it only after scrolling, resizing, or another layout change;
- isolate the drag ghost and insertion marker so their movement does not rerender every project and session row; and
- apply `will-change: transform` only around reorder animation instead of permanently.

The persistence algorithm itself is already appropriate: validating and writing one complete ordered list is linear, simple, and reliable for this data size.

### 2. Extending this UX to Chat Projects

The pure ordering pieces are a good foundation, but the current implementation is not a drop-in Chat Mode feature. Before adding a second consumer:

1. Extract generic ordering primitives from `agentProjectOrdering.ts`, keyed through `getId(item)` rather than the Agent-specific `path` property. Keep filesystem-root merging and Agent session grouping in an Agent-only module.
2. Extract the pointer threshold, capture/cancellation, row measurement, insertion marker, ghost, auto-scroll, and FLIP behavior from `AgentSidebarContent` into a shared hook/component. It should accept an explicit scroll-container ref and an `onCommit(order)` callback, leaving persistence outside the UI primitive.
3. Decide Chat persistence semantics first. Chat Projects are cloud entities, so server-synced ordering is preferable. That likely means a position/rank field plus a reorder endpoint with version or stale-set protection. If ordering is intentionally local-only, use user-scoped browser/app storage rather than the Agent Tauri path store.
4. Centralize the Chat reorder mutation in the shared React Query project cache so `ChatHistoryList`, project pickers, and project detail views all see the same canonical order and can optimistically update/rollback together.
5. Apply ordering before search filtering. During a filtered search, either disable dragging or merge the visible reorder back into the full list while preserving hidden projects. Also coordinate drag with Chat's pull-to-refresh, selection mode, mobile menus, expanded conversations, and project create/delete refetches.
6. Add DOM interaction and accessibility coverage, including a keyboard-accessible way to move projects.

This keeps the current PR focused while giving Chat Mode a shared interaction primitive and the persistence model appropriate for cloud-backed projects.

AnthonyRonning:
This is probably fine for now, I will need to consider where settings / preferences like this live long term. 

I'm also not sure how this would behave on mobile but we don't expose agent mode on mobile yet so it's also fine for now probably.

AnthonyRonning (approved):
I'll likely change the entire "backend" architecture here around projects but this doesn't introduce too much that didn't already exist, and the frontend looks good, so merge whenever it's ready.

Issue #635, opened 2026-07-15
Improve Agent Mode timeline readability


marksftw:
## Summary

Agent Mode shows reasoning and tool activity in a detailed timeline. Improve that timeline in small, independent steps so users can quickly understand what the Agent is doing and scan longer runs without losing access to details.

## Tracked work

- [ ] #646 — Give completed Agent thoughts descriptive labels
- [ ] #647 — Persist descriptive Agent thought labels with task history
- [ ] #648 — Make Agent tool calls compact and easy to recognize
- [ ] #649 — Group Agent reasoning and tool calls into collapsible steps

## Approach

Keep each change focused and preserve existing behavior outside its scope. Start with descriptive labels, then add durable persistence, improve individual tool rows, and finally explore grouping related activity into collapsible steps.

Detailed requirements and acceptance criteria belong in the child issues above.

AnthonyRonning:
I don't think the proposed scenario makes sense because reasoning is separate from tool calling. I don't like the idea of hiding the tool calling inside of thinking as it is streaming in.

I think more thought should go into the streaming look vs finsihed streaming look, tool calling vs content vs reasoning. Because the rebric of scenarios here leaves a lot of edge cases.

marksftw:
I'm fine leaving the tool calling in the main flow for now. I think long term we'll want hidden behind something by default. It ends up being lots of "shell" and "read" on the screen that the user doesn't always need to see.

Here's an example of lots of tool calls where it would be nice to know what's going on:
<img width="474" height="750" alt="Image" src="https://github.com/user-attachments/assets/8b7b1e61-ffec-4b8f-a0b4-2a529a9640f1" />

Often times the "Thought" area has a good explanation why the reasoning decided to launch the tools. I'd like to use that and surface a short version in the main UI flow so the user has more insight into what's going on.

For example, this "Thought" message is almost already good enough. Have gpt-oss make a shorter version and replace the "Thought" text.
<img width="444" height="209" alt="Image" src="https://github.com/user-attachments/assets/e2e9298a-238c-4cf2-b2cd-90fb28dbfa7f" />

AnthonyRonning:
Feel free to try a couple things. I think there's a lot of iterative improvements that it would be easier to do instead of one big sweep but if you find an approach that looks and feels good then I'd be curious to see.

marksftw:
okay, I'll try some steps along the way and see how it goes.

PR #650, opened 2026-07-19
Give completed Agent thoughts descriptive labels

### LINKED ISSUE #646: Give completed Agent thoughts descriptive labels
## Summary

In Agent Mode, completed reasoning rows use the generic label `Thought`. Replace that label with a short description generated by an LLM while leaving the reasoning and the rest of the timeline unchanged.

## Scope

- Keep `Thinking` unchanged while reasoning is streaming.
- When a newly observed reasoning phase finishes, start a non-blocking UI request using `auto:quick`.
- Send only bounded copies of the latest user request from that turn and the completed reasoning text. Do not send tool inputs or outputs, file contents, approvals, the final response, or the full conversation.
- Ask for one short action description. Accept only trimmed, single-line text of 80 characters or fewer.
- If a valid response arrives, replace only the `Thought` label. If the request fails or the response is invalid, silently keep `Thought`.
- Make at most one request per reasoning phase. Loading old history must not generate new labels.
- Store successful labels in `localStorage` under a stable account-, task-, and reasoning-phase-scoped key. Store only the label and identifiers, not the user request or reasoning text.
- Reuse saved labels after switching tasks or restarting the app, and remove them when the related task or Agent history is deleted.
- Keep the implementation in the frontend UI. Do not change Rust, server APIs, or server-side persistence.

## Done when

- A completed live thought changes from `Thought` to its generated description when available.
- Name generation never delays Agent activity, approvals, the final response, or run completion.
- Timeline order, raw reasoning, tools, approvals, and final responses behave exactly as they do today.

## Out of scope

- Persisting or syncing labels with task history
- Redesigning tool-call rows
- Grouping reasoning and tools into a shared container

marksftw:
Closes #646

## What changed

- Track newly completed Agent reasoning phases without changing timeline order or content.
- Generate one bounded, phase-specific present-tense label with `auto:quick`.
- Keep `Thinking` visible while the label is pending, fall back to `Thought` after five seconds or on failure, and fade between label changes.
- Validate and locally persist only single-line labels, reuse cached labels, and clear them with the related task, turn, or Agent history.
- Coalesce duplicate in-flight requests and invalidate pending work safely when history is replaced or deleted.

## Why

Completed reasoning rows previously used the hard-coded `Thought` label, which gave no indication of what a phase had done. The generated-label prompt now treats the user request as background and prioritizes the current phase's latest concrete action, producing more useful labels without adding conversation history, tool data, or an immediate-action field.

## Impact

Label generation is non-blocking and does not delay Agent activity, approvals, final responses, or run completion. Invalid or failed responses remain a silent `Thought` fallback, and loading old history does not trigger new requests.

## Validation

- `bun run test` — 170 passing
- `bun run typecheck`
- `bun run lint` — 0 errors (12 existing warnings)
- `bun run format:check`

The desktop binary was not built.

## Summary by CodeRabbit

- **New Features**
  - Added generated, per–agent thinking phase labels with deterministic phase tracking and timeline-aware rendering.
  - Introduced automatic provisional-to-final label generation and commit flow across active runs.
  - Added reduced-motion-aware “thinking/thought” label UI with animated dots and accessible live updates.
- **Bug Fixes**
  - Improved consistency across completed, failed, cancelled, and restored sessions; obsolete labels are now invalidated after timeline/run changes.
- **Tests**
  - Added/expanded Bun coverage for label generation, timeline phase tracking, markdown label rendering, and run-finish lifecycle behavior.

### DIFF
--- a/frontend/src/components/AgentMode.test.ts
+++ b/frontend/src/components/AgentMode.test.ts
@@ -1,0 +1,279 @@
+import { describe, expect, test } from "bun:test";
+import { handleAgentModeThoughtRunFinished } from "./agent/agentModeThoughtRun";
+import type { AgentEventEnvelope, AgentTimelineItem } from "@/services/agentRuntimeService";
+import {
+  AGENT_THOUGHT_LABEL_PENDING_TEXT,
+  AGENT_THOUGHT_LABEL_PROVISIONAL_DELAY_MS,
+  AGENT_THOUGHT_LABEL_PROVISIONAL_MILESTONES_MS,
+  AgentThoughtLabelFinalRequestRegistry,
+  AgentThoughtLabelProvisionalScheduler,
+  startAgentThoughtLabelDisplay
+} from "@/services/agentThoughtLabels";
+import { AgentLiveThoughtPhaseTracker, type AgentThoughtPhase } from "@/services/agentTimeline";
+
+const SESSION_ID = "session";
+const PROVISIONAL_LABEL = "Investigating login flow";
+const REASONING =
+  "Evaluating the login flow and tracing session state across runtime events before choosing the safest correction ".repeat(
+    2
+  );
+
+function deferred<T>(): { promise: Promise<T>; resolve: (value: T) => void } {
+  let resolve: ((value: T) => void) | undefined;
+  const promise = new Promise<T>((nextResolve) => {
+    resolve = nextResolve;
+  });
+  return { promise, resolve: (value) => resolve?.(value) };
+}
+
+function manualTimers(): {
+  schedule: (callback: () => void, delayMs: number) => () => void;
+  run: (delayMs: number) => void;
+} {
+  const timers: Array<{ callback: () => void; delayMs: number; cancelled: boolean }> = [];
+  return {
+    schedule: (callback, delayMs) => {
+      const timer = { callback, delayMs, cancelled: false };
+      timers.push(timer);
+      return () => {
+        timer.cancelled = true;
+      };
+    },
+    run: (delayMs) => {
+      timers
+        .filter((timer) => !timer.cancelled && timer.delayMs === delayMs)
+        .forEach((timer) => {
+          timer.cancelled = true;
+          timer.callback();
+        });
+    }
+  };
+}
+
+function timeline(reasoningText = REASONING): AgentTimelineItem[] {
+  return [
+    {
+      id: "user",
+      itemType: "message",
+      role: "user",
+      text: "Inspect login",
+      createdMs: 0,
+      merge: "replace"
+    },
+    {
+      id: "thought",
+      itemType: "thinking",
+      role: "thought",
+      text: reasoningText,
+      createdMs: 1,
+      merge: "replace"
+    }
+  ];
+}
+
+function liveTracker(items = timeline()): AgentLiveThoughtPhaseTracker {
+  const tracker = new AgentLiveThoughtPhaseTracker();
+  items.forEach((item) => tracker.observeTimelineItem(SESSION_ID, item));
+  return tracker;
+}
+
+async function flushPromises(): Promise<void> {
+  await Promise.resolve();
+  await Promise.resolve();
+}
+
+describe("AgentMode runFinished thought labels", () => {
+  for (const status of ["completed", "failed"] as const) {
+    test(`${status} preserves its provisional through the authoritative reload`, async () => {
+      const authoritativeTimeline = timeline();
+      const tracker = liveTracker(authoritativeTimeline);
+      const timers = manualTimers();
+      const finalResponse = deferred<string | null>();
+      const finalRegistry = new AgentThoughtLabelFinalRequestRegistry();
+      const visibleLabels: Array<string | null> = [AGENT_THOUGHT_LABEL_PENDING_TEXT];
+      let visibleLabel: string | null = AGENT_THOUGHT_LABEL_PENDING_TEXT;
+      let finalRequestCount = 0;
+      let replaceCount = 0;
+      const setVisibleLabel = (label: string | null) => {
+        if (visibleLabel === label) return;
+        visibleLabel = label;
+        visibleLabels.push(label);
+      };
+      const provisionalScheduler = new AgentThoughtLabelProvisionalScheduler({
+        schedule: timers.schedule,
+        request: async () => PROVISIONAL_LABEL,
+        commit: (_phase, label) => setVisibleLabel(label)
+      });
+      provisionalScheduler.observe(tracker.activePhase(SESSION_ID)!);
+      timers.run(AGENT_THOUGHT_LABEL_PROVISIONAL_DELAY_MS);
+      await flushPromises();
+      expect(visibleLabel).toBe(PROVISIONAL_LABEL);
+
+      const finalizePhase = (phase: AgentThoughtPhase) => {
+        const retainedLabel = provisionalScheduler.complete(phase.sessionId, phase.phaseId);
+        const finalRequest = finalRegistry.begin(phase, retainedLabel);
+        if (!finalRequest) return;
+        finalRequestCount += 1;
+        const cancel = startAgentThoughtLabelDisplay({
+          retainedLabel: finalRequest.retainedLabel,
+          request: () => finalResponse.promise.finally(finalRequest.finish),
+          commit: (label, expectedLabel) => {
+            if (!finalRequest.isCurrent()) return;
+            if (expectedLabel !== undefined && visibleLabel !== expectedLabel) return;
+            finalRequest.recordLabel(label);
+            setVisibleLabel(label);
+          }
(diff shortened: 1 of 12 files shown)


AnthonyRonning:
Screenshot 2026-07-19 at 9 49 28 PM

Just tested it, had a few thoughts on the approach. 

When it comes to the thinking thoughts, I know there's a couple approaches. One is a continual summarizer that runs on whatever tokens have already streamed in. Updating after a few seconds with the latest amount. This gives an appearance like it is actively thinking. 

Another is sort of combining the thoughts + content + tool calls + repeat into the collapsed form and doing something similar with the continuous summarizer. Though I kinda like that they stream in first, and then collapse down.  

Separately of the approach, I think whatever prompt you're using for the summarization would need some more examples of what it should output. These make it look like it is making the tool calls, but it is just thinking about the tool calls it might want to make.

We should use llama or gemma for this, not gpt-oss, It has random problems at times.

marksftw:
Thanks, this makes sense. I agree the current labels can sometimes sound like the agent is already making a tool call when it is really only considering one. I'll adjust that.

As for the concept of collapsing the reasoning and tool calling into UI phases, that is already part of the broader [[Agent timeline exploration](https://github.com/OpenSecretCloud/Maple/issues/635)](https://github.com/OpenSecretCloud/Maple/issues/635). I'm trying your suggestion from last week to break UX experiments into smaller chunks instead of trying to redesign the whole timeline at once. I also like the direction of letting activity stream in first and collapsing it afterward.

For this pass, I’m going to try gemma, strengthen the prompt with examples that distinguish intent from actual execution, and test a bounded version of the continual summarizer: one provisional label during longer thoughts, followed by a final label when the thought ends. If gemma isn’t fast enough, I’ll try llama and retune the prompt for it.

AnthonyRonning:
Makes sense. Make sure to set reasoning off for gemma. I have an example of that with my auto approver classifier for tool calls.

marksftw:
The direction here is to make Agent activity feel more legible without turning this PR into a full timeline redesign. Instead of leaving someone staring at a generic “Thinking” or “Thought” label, the UI provides a short description of the current reasoning phase. This remains a focused step within the broader [[Agent timeline exploration](app://-/index.html#635)](#635).

The latest update:

- Uses Gemma with model reasoning disabled, plus a stronger prompt and examples that favor concrete subjects and distinguish considering an action from actually performing it.
- Asks for labels between 3 and 8 words. That range is controlled in the prompt and can be adjusted in code.
- Generates bounded provisional labels during longer reasoning phases at the 1-, 5-, and 15-second milestones. Requests have a short deadline, skip unchanged reasoning, and are capped to avoid excessive background work.
- Generates a final label when the reasoning phase completes. Any provisional label remains visible while that request settles, so the UI does not jump backward to “Thinking.”
- Keeps the animated “Thinking…” state until a descriptive label is available. “Thought” remains the fallback when label generation fails or does not return in time.
- Saves displayed labels in localStorage so they remain stable when returning to a task. These labels are not persisted to the server.
- Fades between label changes and cancels stale requests when phases complete, tasks are replaced or deleted, or account history is cleared.

AnthonyRonning:
Thanks for the follow-ups. The direction is improved, and the Gemma choice, disabled reasoning, provisional labels, and expanded intent-versus-execution examples all address the feedback in code. Before I approve this, I would like the following changes.

1. Validate and tune the prompt against the real Gemma model.

The current tests verify that prompt text and examples exist, then feed mocked labels back. They do not show that Gemma actually follows the distinction. In a live test, the reasoning said “Let me read both files…” and “I’ll read both files…”, but the generated label was “Reading agentThoughtLabels and package.json.” Read tool calls did follow, so the complete timeline was understandable, but the reasoning label still converted a prospective action into active execution.

Please spend time prompt-engineering this with the actual model rather than inferring that the prompt should work. Build a small fixed evaluation set covering planning, considering a tool, actively executing a tool, interpreting results, and producing a final response, and test the real outputs. Since reasoning text is not authoritative tool state, the safest initial rule may be to reserve verbs such as Reading, Running, Searching, and Testing for labels derived from actual tool events. Reasoning-only labels can use Planning, Evaluating, Comparing, or similar language.

2. Make the label state machine race-safe.

An older provisional request can currently finish after newer reasoning arrives and commit a label based on the obsolete snapshot. That older label can then become the durable fallback if final generation fails. An older thought summary must never overwrite a newer one.

Please associate every request with the exact phase/snapshot generation that started it, discard any result that is no longer current, and abort the previous request whenever a newer snapshot or newer thought supersedes it. The final-label request also needs an AbortSignal. Completion, cancellation, task deletion, history replacement, unmount, and a newer thought beginning should terminate any request that is no longer useful. Please add lifecycle tests that deliberately resolve requests out of order and prove that only the newest valid result can commit.

3. Preserve provisional labels when a run fails.

The failed-run path currently discards the provisional scheduler entry. The subsequent authoritative reload reconstructs the phase without the retained label, producing a provisional → Thinking… → final/fallback regression on an already-terminal run. Failed runs should preserve and finish the active thought in the same way as completed runs. Only an explicitly cancelled run should discard it. Please cover completed, failed, and cancelled sequencing in AgentMode-level tests.

4. Keep thought summaries in memory only for this first version.

Please remove localStorage persistence and the positional v1 cache. We do not need to recover generated summaries after restarting the application or when opening older tasks. If a summary is not already available in current in-memory state, a completed historical reasoning block should simply show the fallback “Thought.”

This also avoids establishing thought-N plus cached labels as a persistence contract while the Goose timeline representation is still evolving. No source hash, prompt migration, or recovery scheme is needed for this version because the generated metadata should be disposable.

The normal success path and background-run recovery are moving in a good direction. The remaining work is to prove the model behavior and make concurrent request/lifecycle transitions deterministic before merging.

marksftw:
Thanks for the detailed review — this was really helpful. I pushed a follow-up that addresses each of these areas.

- **Gemma prompt validation:** I built a fixed 11-case evaluation set covering planning, prospective/claimed tool use, interpreting results, comparisons, and final responses, and ran each case three times against the real Gemma model through Maple. The baseline reproduced the reported “I’ll read…” → “Reading…” problem 3/3 times. After tuning, all 33 runs produced accepted labels, with no reserved tool-execution verbs or missed concrete subjects. Gemma reasoning remains disabled, and reasoning-only labels now reject execution-leading verbs such as Reading, Running, Searching, and Testing. The live evaluation harness was development-only and is not included in the app.

- **Race-safe state:** Provisional and final requests now belong to the exact phase snapshot that started them. Newer snapshots abort older work, stale results cannot commit, and final requests receive an AbortSignal. Cancellation, deletion, history replacement, unmount, and superseding thoughts all invalidate requests that are no longer useful. The tests deliberately resolve requests out of order to verify that only the newest valid result wins.

- **Failed-run behavior:** Completed and failed runs now both preserve the provisional label and finish the active thought; only an explicitly cancelled run discards it. I added an AgentMode event-boundary integration test that exercises completed, failed, and cancelled `runFinished` events through the authoritative timeline reload. It verifies that failed runs never regress from a provisional label to Thinking/Thought, reloads do not start duplicate final requests, and cancelled work cannot commit late.

- **Memory-only labels:** All thought-label localStorage persistence and recovery code has been removed. Labels live only in the current in-memory AgentMode state, and completed historical thoughts without an in-memory label simply render the `Thought` fallback.

The frontend suite is green with 196 tests, along with typecheck and formatting; lint has no errors.

marksftw:
Rebased this branch onto the latest master so the PR includes the recent trusted-local-skills and proxy updates. The only conflict was in AgentMode.tsx; I resolved it by preserving both the upstream skills-trust guard and the thought-label lifecycle. I reran the frontend checks afterward: 205 tests, typecheck, formatting, and lint all pass.

AnthonyRonning:
One remaining merge blocker after the rebase: the frontend suite is still red both locally and in GitHub CI (203 passed, 2 failed), at frontend/src/services/agentThoughtLabels.test.ts:328 and :397. This looks like test synchronization rather than a production behavior failure: both requests are finally-chained and the assertions only flush one microtask. Replacing those await Promise.resolve() calls with the existing await nextTask() helper should let the chained settle/commit finish. Please fix those two tests, rerun the suite, and update the green-test claim. I do not have another blocker for the proof-of-concept scope.

marksftw:
Fixed both test synchronization points in 574f16e. The assertions now use the existing nextTask() helper so the finally-to-then chain settles before results are checked; this is test-only and does not change production behavior. I reran the full frontend CI sequence with Bun 1.3.5: formatting, lint (0 errors), typecheck, and all 205 tests pass. Thanks for catching the runtime-specific timing difference.

AnthonyRonning (approved):
Approved. The test synchronization fix is the correct scoped change, the full CI matrix is green, and I have no remaining blockers for this proof of concept. The branch is currently clean against master; please rebase if master advances before merge, then squash merge when ready.

PR #665, opened 2026-07-21
Stabilize provisional thought labels during streaming

### LINKED ISSUE #660: Allow provisional thought labels during continuous reasoning
Follow-up to #646 and the implementation in #650.

## Problem

Every streamed reasoning update currently supersedes the active snapshot and aborts its provisional Gemma request. During a long, continuously updating thought, the 1-, 5-, and 15-second attempts can all be cancelled, leaving Thinking... visible until final label generation begins after the thought ends.

This came from the important race-safety requirement that a label from an older reasoning snapshot must not arrive late, overwrite a newer summary, or become the retained fallback.

## Proposed compromise

Treat the bounded 1-, 5-, and 15-second samples as the authoritative provisional generations. Buffer intervening reasoning for the next sample instead of aborting the current request on every streamed delta. A newer sample, newer thought, terminal transition, deletion, history replacement, or unmount still aborts and invalidates obsolete work, and generation guards ensure only the newest valid request can commit.

Add continuous-stream and out-of-order lifecycle tests for this behavior.

marksftw:
Closes #660

Follow-up to #646 and #650.

## Why

Long active reasoning phases could remain on `Thinking...` because every streamed reasoning delta invalidated and aborted the in-flight Gemma request before it could return.

That strict behavior came from a valid review concern: a result tied to older reasoning must not arrive out of order, overwrite a newer summary, or become the retained fallback. However, treating every transport update as a new authoritative snapshot made provisional labels effectively impossible during continuous streaming. Allowing every timed result to publish solved that starvation but made the label change too often to read comfortably.

This PR uses bounded sampled generations as the compromise: streaming text continues to accumulate, while only deliberate label checkpoints supersede an in-flight request.

## What changed

- Keep live reasoning updates buffered without cancelling the current sampled request.
- Try an initial provisional label after one second and use the five-second checkpoint only when no provisional has succeeded.
- Allow one intentional refresh after fifteen seconds when the reasoning has changed, then let the completed-phase label remain authoritative.
- Abort an older request when a newer sampled generation, task context, thought phase, or lifecycle transition supersedes it.
- Guard every commit by phase, generation, controller, and sampled content so late results cannot overwrite newer work.
- Retain an accepted provisional if completion cancels a pending fifteen-second refresh.

The usual visible progression is now `Thinking...` → early provisional → optional fifteen-second refresh → final label. Prompting, model selection, persistence, privacy boundaries, and label validation are unchanged.

## Validation

- `bun test` — 206 passing
- `bun run typecheck`
- `bun run lint` — 0 errors, 12 existing warnings
- `bun run format:check`

The desktop binary was not built.

### DIFF
--- a/frontend/src/components/AgentMode.test.ts
+++ b/frontend/src/components/AgentMode.test.ts
@@ -214,6 +214,8 @@ describe("AgentMode runFinished thought labels", () => {
     });
     provisionalScheduler.observe(tracker.activePhase(SESSION_ID)!);
     timers.run(AGENT_THOUGHT_LABEL_PROVISIONAL_MILESTONES_MS[1]);
+    expect(provisionalSignals).toHaveLength(1);
+    timers.run(AGENT_THOUGHT_LABEL_PROVISIONAL_MILESTONES_MS[2]);
     expect(provisionalSignals).toHaveLength(2);
 
     const finished = handleAgentModeThoughtRunFinished({
--- a/frontend/src/services/agentThoughtLabels.test.ts
+++ b/frontend/src/services/agentThoughtLabels.test.ts
@@ -450,15 +450,16 @@ describe("AgentThoughtLabelProvisionalScheduler", () => {
     await Promise.resolve();
   });
 
-  test("refreshes from the latest reasoning at one, five, and fifteen seconds", async () => {
+  test("refreshes from the latest reasoning until a provisional succeeds", async () => {
     const timers = manualTimers();
     const requestedReasoning: string[] = [];
     const commits: string[] = [];
+    const results = [null, null, "Reviewing latest login evidence"];
     const scheduler = new AgentThoughtLabelProvisionalScheduler({
       schedule: timers.schedule,
       request: async (source) => {
         requestedReasoning.push(source.reasoningText);
-        return `Label ${requestedReasoning.length}`;
+        return results.shift() ?? null;
       },
       commit: (_source, label) => commits.push(label)
     });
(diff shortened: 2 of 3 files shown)


marksftw:
Post-rebase code review is complete. I reviewed the change independently from race/lifecycle, simplicity/privacy, and UX/test-coverage angles, then validated the findings locally. No code changes were warranted.

The snapshot/request guards keep older results from overwriting newer labels, lifecycle cleanup remains deterministic, and the implementation stays narrowly scoped without reintroducing persistence or development diagnostics. I also checked the early-failure/no-retry behavior and synthetic same-phase replacement behavior; both are consistent with the intended lifecycle and do not justify additional complexity in this small follow-up.

Validation is clean: 214 tests passed, typecheck and formatting passed, lint reported 0 errors (12 existing warnings), and `git diff --check` passed.

marksftw:
Ready for review. Tested and feeling good about the UX for this. Labels are generated quickly. There is a 5 second retry if it fails. And then longer reasoning blocks get an updated label at 15 seconds because they have more understanding by that point.

AnthonyRonning (approved):
Approved. This is a focused, maintainable follow-up that stabilizes provisional thought labels during streaming, including lifecycle and cancellation handling, without adding persistence or unrelated scope. The review found no blocking issues. One minor retry edge case—unchanged reasoning after an early label request failure—can be handled later; it only leaves the generic “Thinking…” state and does not risk stale or incorrect labels. You can merge whenever you are ready.

PR #677, opened 2026-07-23
tweaks: tool display reporting and markdown

michaelneale:
Tools show up like this: 

image

in agent mode

and I also tried to tweak the markdown handling so it doesn't switch to it always with indentation. I found in non agentic cases (well others did) sometimes GLM in particular would return blocks of text which would be rendered incorrectly as code blocks (this one is harder to validate, but thought it may help).

### DIFF
--- a/frontend/src-tauri/src/agent.rs
+++ b/frontend/src-tauri/src/agent.rs
@@ -3120,7 +3120,10 @@ fn message_to_timeline_items_with_thinking(
                 id: format!("permission-{}", request.id),
                 item_type: "permission".to_string(),
                 role: Some("system".to_string()),
-                title: Some(format_tool_title(&request.tool_name)),
+                title: Some(
+                    descriptive_tool_title(&request.tool_name, &request.arguments)
+                        .unwrap_or_else(|| format_tool_title(&request.tool_name)),
+                ),
                 text: request.prompt.clone(),
                 status: Some("pending".to_string()),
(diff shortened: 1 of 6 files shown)


frontend/src/components/UnifiedChat.tsx
AnthonyRonning, line 1: UnifiedChat is the main chat interface as it currently exists. Agent mode doesn't use this and the tool components are seperate right now, intentionally for feature flag reasons and to not change our main interface for now. We'll do a consolidation soon once we have a tool display we like that works well for both. For now just revert UnifiedChat changes.

frontend/src-tauri/src/agent.rs
AnthonyRonning, line 1: Now that I actually read the code in this file, I think i'll refactor it at some point soon to move more of the display logic to the frontend side but for now this follows existing standards so I won't block on it now.

AnthonyRonning (changes requested):
Reverting UnifiedChat is the only change I'd request here, let's keep this change for agent mode only. 

The markdown fix is a good catch, I have always wondered why kimi was doing that, had no idea it was due to 4 spaces.

frontend/src/components/UnifiedChat.tsx
michaelneale: It was non agent mode where this was biting - so was deliberate but can move that to separate thing

AnthonyRonning (approved):

Issue #654, opened 2026-07-20
Harden the credential-backed Local Proxy against arbitrary browser origins


AnthonyRonning:
## Context

The desktop Local Proxy currently permits wildcard browser origins when CORS is enabled. The proxy also falls back to Maple's saved credential when an incoming request does not provide its own Authorization header.

That combination creates a billing and security boundary worth treating explicitly: a browser origin that can reach the loopback endpoint under the browser's local-network policy may be able to make inference requests using the saved credential.

## Scope

- Confirm behavior across supported browsers and their local-network protections.
- Decide whether CORS should remain enabled by default for new configurations.
- Define migration or consent behavior for existing saved configurations.
- Consider origin restrictions or requiring an explicit inbound credential whenever CORS is enabled.
- Keep user-facing warnings and setup guidance aligned with the chosen policy.

This should be designed and reviewed separately from PR #652, which should remain a documentation and presentation change.

PR #711, opened 2026-07-28
Fence manual proxy startup across account transitions

### LINKED ISSUE #655: Fence manual proxy startup against logout and account transitions
## Problem

First-time manual proxy startup creates an API key before entering the serialized proxy operation. If API-key creation or the following key-list refresh is delayed while the user logs out, logout can reset the proxy first and the stale startup handler can enqueue a new start afterward using the prior account's key.

The component-level account guard explored in PR #652 would only partially mitigate this: it can leave a newly created remote key orphaned and does not move the account boundary into the service lifecycle.

## Expected behavior

A manual proxy setup started by one authenticated account must not persist or start after that account logs out or another account becomes active.

## Suggested direction

- Fence manual proxy startup at the lifecycle or service boundary using the initiating account and an auth generation.
- Coordinate key creation with logout cleanup rather than guarding only the React component.
- Revoke or otherwise roll back a key created by an operation that becomes stale.
- Add a deterministic test using delayed key creation followed by logout and account transition.

Track and fix this separately from the documentation and UI work in PR #652.

AnthonyRonning:
## Summary

- capture the initiating account and an in-memory auth generation before first-time manual API-key creation
- reject delayed work before native startup when logout or an account transition changes that generation
- invalidate in-flight startup synchronously when logout cleanup begins, and scrub a proxy if native startup completes after the account changed

## Backward compatibility

- no proxy config schema, native persistence, Credential Manager, auto-start, migration, or login-time ownership behavior changed
- released ownerless saved-key and auto-start behavior remains exactly the same as master
- normal existing-key and first-time key creation flows keep their prior UI and saved-config behavior
- this PR remains scoped to #655 and does not include #654

## Validation

- focused proxy lifecycle tests: 8 passed
- full frontend tests: 389 passed, 0 failed, 1,351 assertions
- frontend format, typecheck, production build, and lint: 0 errors; 13 pre-existing warnings
- Rust: 213 passed, 0 failed, 1 expected OCR test ignored; cargo fmt check passed
- managed workspace smoke passed with OpenSecret, Postgres, Continuum feature flags, billing, and billing Postgres
- rebuilt isolated macOS GUI: released-format ownerless config cold-started while signed out, retained the same saved key and auto-start setting after Pro login, showed Proxy running / Online, and returned 10 models; normal logout then stopped and scrubbed it exactly as master does
- three fresh-context final reviews found no high or critical blocker

Closes #655

### DIFF
--- a/frontend/src-tauri/src/lib.rs
+++ b/frontend/src-tauri/src/lib.rs
@@ -137,6 +137,21 @@ pub fn run() {
             tts::tts_delete_models,
         ])
         .setup(|app| {
+            // Released proxy configs have no owner metadata. Preserve their
+            // existing auto-start behavior until an explicit authenticated
+            // start/save opts the config into account fencing.
+            {
+                let app_handle_proxy = app.handle().clone();
+                tauri::async_runtime::spawn(async move {
+                    tokio::time::sleep(tokio::time::Duration::from_millis(500)).await;
+                    if let Err(error) =
+                        proxy::init_ownerless_proxy_on_startup(app_handle_proxy).await
+                    {
+                        log::error!("Failed to initialize existing proxy config: {error}");
+                    }
+                });
+            }
+
             // Set up the deep link handler
             // Use a cloned handle with 'static lifetime
             let app_handle = app.handle().clone();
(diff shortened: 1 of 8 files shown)


AnthonyRonning:
Closing without merge. After implementation and backward-compatibility validation, we decided this hypothetical in-flight startup race does not justify the added lifecycle complexity. Existing logout already stops and scrubs the local proxy. See #655.

PR #708, opened 2026-07-28
Fix later chat history pagination during rapid upward scrolling

AnthonyRonning:
## Summary

- recheck the history boundary after native upward scrolling applies for wheel, touch, keyboard, and scrollbar input
- synchronize the comparison baseline after anchor restoration so programmatic scrolls cannot become fresh pagination intent
- add focused coverage for the post-scroll retry path and forward/stationary movement

## Validation

- `bun run format:check`
- `bun run lint` (0 errors; 13 existing warnings)
- `bun test` (361 passed)
- `bun run build`
- request-instrumented web GUI with a 90-item tool-heavy chat: exactly one older-page request per deliberate reach through pages 2-5, anchor drift within 0.25 px, and no request past the oldest item
- isolated native macOS WKWebView cold-start smoke through later pages, plus short-chat top-alignment regression check
- three independent reviews (correctness, lifecycle/concurrency, and UX/regression) found no high- or critical-severity defects

Closes #692

PR #713, opened 2026-07-28
Fix rapid later chat history pagination

### LINKED ISSUE #692: Make later Chat history pagination reliable during rapid upward scrolling
## Summary

Follow up after #687 on the reliability of second-and-later Chat history pages during rapid upward scrolling.

Manual testing of #687 is significantly better and no longer merge-blocking. However, when scrolling upward quickly, a later page can occasionally feel stuck: reaching the top does not fetch the next page until the user retries.

## Desired behavior

After each prepend settles, one deliberate upward reach of the real top boundary should reliably load exactly one older page, including pages 3 and 4, without requiring extra down/up gestures.

## Guardrails

- Do not fetch older history on mount or merely because an initial page underfills the viewport.
- Do not allow idle or momentum to cascade through multiple pages.
- Preserve the visible anchor across every prepend.
- Keep short chats top-aligned.
- Preserve the compact tool-card presentation and the correctness improvements from #687.

## Validation

- Reproduce with a tool-heavy conversation spanning several history pages and rapid upward scrolling.
- Verify pages 2–4 each load once from one deliberate top reach, with no retry.
- Verify idle/momentum issues no extra request and the visual anchor remains stable.

Related: #687

AnthonyRonning:
## Summary

- end the consumed wheel gesture synchronously when event timestamps show the existing 180 ms quiet boundary, even if a busy prepend delayed the fallback timer
- retain the one-page-per-gesture and load-in-flight guards
- add focused coverage for delayed timer callbacks and rapid pages 2 through 4

The failure on master was reproduced with a 296 ms gap between consecutive upward wheel events: the first prepend delayed the gesture timeout, then the second wheel event cleared that still-pending timeout and stayed attached to the already-consumed gesture.

## Validation

- native macOS Maple app against the managed local OpenSecret, billing, feature-flag, and database services
- tool-heavy 65-item conversation spanning four history pages
- rapid later-page gestures loaded exactly one page each, preserved the visible anchor, and issued no idle cascade
- final cold runtime recorded three advancing older-page requests total: the initial older page plus two back-to-back later pages
- bun test: 387 passed
- bun run typecheck
- bun run lint: 0 errors, 13 existing warnings
- bun run build
- managed workspace smoke
- fresh correctness, lifecycle/security, and UX/regression reviews

Supersedes #708.

Closes #692

### DIFF
--- a/frontend/src/components/UnifiedChat.tsx
+++ b/frontend/src/components/UnifiedChat.tsx
@@ -81,6 +81,7 @@ import { isLinux, isTauri } from "@/utils/platform";
 import { ConversationProjectPicker } from "@/components/ConversationProjectPicker";
 import {
   CHAT_HISTORY_TOP_MARGIN_PX,
+  CHAT_HISTORY_WHEEL_GESTURE_QUIET_MS,
   ChatHistoryPaginationGate,
   chatHistoryCursorProgressed,
   requiredChatHistoryBottomCompensation,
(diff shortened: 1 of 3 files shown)


AnthonyRonning:
Anthony's manual test found no behavioral difference. My own cold exact-sequence evidence also showed that the second gesture could still be swallowed when it arrived during the first prepend/loading transition; I incorrectly treated that as out of scope and validated after resetting the scroll state. Closing this PR because it does not fix #692's actual interaction.

PR #715, opened 2026-07-29
fix(auth): keep OAuth callback mounted on login

AnthonyRonning:
## Summary

- keep the OAuth callback route mounted when authentication changes from signed out to a user
- retain account-keyed remounts for ordinary routes and global account-sensitive UI
- keep the persistent chat runtime account-keyed and scoped to authenticated Home content
- add regression coverage for exactly-once callback processing, cross-account resets, and same-account chat persistence

## Why

A successful OAuth callback updates auth.user. The account-keyed chat provider previously wrapped the router outlet, so that update remounted the callback route and reset its component-local processed guard. Maple then submitted the same code and state again. A consume-once backend correctly rejects that replay, which caused the user to briefly see a login error after the first request had already succeeded.

This change is frontend-only. It does not change SDK behavior, callback payloads, backend status codes, token storage, or native handoff contracts. It is compatible with both older backends and the consume-once OAuth state behavior in OpenSecretCloud/opensecret#222.

## Validation

- frontend tests: 388 passed
- focused regression tests: passed
- TypeScript typecheck: passed
- production build: passed
- formatting: passed
- lint: 0 errors; 13 pre-existing warnings
- two fresh lifecycle/security reviews: no actionable findings

A live provider smoke test remains recommended after deploying the hosted frontend.


AnthonyRonning:
Related frontend hardening is intentionally split into #716: Apple popup authentication had its own duplicate-completion path (DOM event plus signIn promise). The two PRs are independently reviewable and neither changes backend or SDK error semantics.

PR #744, opened 2026-08-08
Align new chat and new task screen layouts

### LINKED ISSUE #743: Make the New Chat and New Task screen have similar layout
1. The words "Research Anything..." and "Work on anything..." should have the same vertical placement on the screen and everything should flow correctly below it. If a user clicks between Chat and Agent, those two text labels should be on the same Y-axis. That means that only the New Task screen needs to change so it more closely resembles the New Chat screen.
2. For both new chat and new task, the input box is too wide. It should still dynamically change width based on window size, but with a max width of 650px. Right now it looks like the max width is around 850px. This only applies to the standard input box on the "new" screen. it does not affect the input box on an existing chat/task. And it does not affect the expanded input box on the new screen. Expanded box keeps its existing behavior.

Screenshots of current app behavior:
Image
Image

marksftw:
## Summary

- align the empty New Task heading and composer with the New Chat vertical layout
- cap the width of New Chat and New Task input box at 650px while keeping them fluid on narrower screens
- preserve existing conversation/task layouts and expanded composer widths

https://github.com/user-attachments/assets/2e1c3826-515b-4763-8a64-d444a272c3fb

## Root cause

New Task centered its empty state inside a fixed 52vh region and inherited the wider `max-w-4xl` container, while New Chat used the remaining full-height content area. The standard new-screen composers also had wider maximum widths than the specification.

## Validation

- pre-commit hook: Prettier, TypeScript, Vite production build, and full Bun test suite
- `519 pass`, `0 fail`
- targeted ESLint
- `git diff --check`

Closes #743

### DIFF
--- a/frontend/src/components/AgentMode.tsx
+++ b/frontend/src/components/AgentMode.tsx
@@ -3594,6 +3594,7 @@ function AgentComposer({
         className={cn(
           CHAT_COMPOSER_TEXTAREA_CLASS,
           onToggleExpanded && "pr-8",
+          onToggleExpanded && !isExpanded && "landscape-short:min-h-[52px]",
           isExpanded &&
             "min-h-0 max-h-none flex-1 overflow-y-auto landscape-short:min-h-0 landscape-short:max-h-none"
         )}
--- a/frontend/src/components/AgentMode.tsx
+++ b/frontend/src/components/AgentMode.tsx
@@ -2632,10 +2632,12 @@ export function AgentMode({ userId }: { userId: string }) {
           >
             <div
               className={cn(
-                "mx-auto w-full p-4 md:p-6 landscape-short:p-2",
-                timelineItems.length === 0 && isAgentFullscreen
-                  ? "flex min-h-full max-w-6xl flex-col"
-                  : "max-w-4xl"
+                "mx-auto w-full",
+                timelineItems.length > 0
+                  ? "max-w-4xl p-4 md:p-6 landscape-short:p-2"
+                  : isAgentFullscreen
+                    ? "flex min-h-full max-w-6xl flex-col p-4 md:p-6 landscape-short:p-2"
+                    : "flex min-h-full flex-col px-4"
               )}
             >
               {timelineItems.length === 0 ? (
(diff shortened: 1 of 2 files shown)


marksftw:
## Independent review follow-up

I ran three independent review passes over this PR from different angles:

1. **Specification fidelity:** compared the complete diff and surrounding layout code against every requirement in #743.
2. **Minimality and invasiveness:** checked whether the state branching, utility classes, and component changes were the smallest clear solution.
3. **Responsive and regression risk:** traced flex sizing, padding, composer controls, breakpoints, short-landscape behavior, and the standard/existing/expanded states independently.

### Reconciled findings

No major issue or unnecessary refactor was found. The main implementation is appropriately scoped:

- Standard New Chat and New Task composers remain fluid and are capped at 650px.
- Their empty-state heading geometry now uses matching full-height centering, spacer, gap, margin, typography, and control-row sizing.
- Existing chats and tasks retain their prior `max-w-4xl` layouts.
- Expanded New Chat and New Task retain their prior `max-w-6xl` behavior.
- The Agent send-button sizing change is restricted to the collapsed new-task composer; existing and expanded task composers are unaffected.
- The overall change remains limited to two component files, with no new abstraction, code comments, or low-value tests.

Two reviewers independently identified one minor edge discrepancy in the first revision: under the `landscape-short` breakpoint, New Task inherited a 40px textarea minimum while New Chat remained 52px. Because both complete stacks are centered, that could place “Work on anything...” roughly 6px below “Research anything...”.

Commit `321646a` fixes that discrepancy with one collapsed-new-task-only `landscape-short:min-h-[52px]` override. I then revalidated Tailwind class resolution for every task state:

- Collapsed New Task resolves to `landscape-short:min-h-[52px]`.
- Existing Task retains `landscape-short:min-h-[40px]`.
- Expanded New Task retains `landscape-short:min-h-0`.

This closes the only verified acceptance gap without broadening the change.

### Validation

- Prettier: passed
- Targeted ESLint: passed
- TypeScript `tsc -b`: passed
- Vite production build: passed
- Full Bun test suite: **519 passed, 0 failed**
- `git diff --check`: passed
- Compiled CSS inspection: confirmed the intended 52px short-landscape utility is emitted
- Final independent re-review of the one-line fix: no regression found

A controllable browser session was unavailable for a fresh bounding-rectangle measurement, so the Y-axis result was validated through the exact flex geometry, compiled responsive CSS, and Tailwind merge behavior. The local app was reachable, and there is no remaining source-level or build-level issue worth expanding the patch to address.

**Conclusion:** the implementation satisfies #743 with a minimal, state-scoped change and no unnecessary complexity.

AnthonyRonning (approved):
looks good

PR #790, opened 2026-08-18
fix(agent): retry transient inference failures for ~3 minutes

AnthonyRonning:
## Summary

- raise Maple Agent Mode's transient inference retry budget from Goose's default **3** extra attempts (~7s) to **10** (~3 minutes)
- keep the existing 1s × 2 backoff, 30s cap, `transient_only` policy, and Cancel-interrupts-backoff behavior
- do not retry deterministic 4xx / client failures, and do not retry after the first successful streamed item

A glm-5-2 500 from OpenSecret/Tinfoil already exhausted the old 4-call / ~7s budget as four immediate `http_status_500` lines, then surfaced `Server error: Maple's server returned status 500`. Long autonomous Agent runs should survive a short provider brownout. No UI attempt counter in this PR; Cancel remains the escape hatch.

## Behavior and scope

- change is confined to `MapleProvider::retry_config()` and the shared-budget unit test
- no Goose fork, no Goose pin bump, no frontend or permission changes
- after this lands, a 500 before the first streamed item can take up to ~3 minutes of silent retry while the run stays active and Cancel stays available

## Validation

- `cargo test --locked "agent::provider::tests::"`: 30 passed
- pre-commit hook ran Prettier, the frontend production build, TypeScript, and the locked Rust suite: **313 passed**, 0 failed, 2 ignored
- not run: native desktop smoke of a live 3-minute 500 streak (we have not reproduced one since local OpenSecret logging was fixed)

### DIFF
--- a/frontend/src-tauri/src/agent/provider.rs
+++ b/frontend/src-tauri/src/agent/provider.rs
@@ -11,7 +11,10 @@ use goose_providers::http_status::is_context_length_exceeded_message;
 use goose_providers::images::ImageFormat;
 use goose_providers::model::ModelConfig;
 use goose_providers::request_log::{start_log, LoggerHandleExt};
-use goose_providers::retry::{should_retry, RetryConfig};
+use goose_providers::retry::{
+    should_retry, RetryConfig, DEFAULT_BACKOFF_MULTIPLIER, DEFAULT_INITIAL_RETRY_INTERVAL_MS,
+    DEFAULT_MAX_RETRY_INTERVAL_MS,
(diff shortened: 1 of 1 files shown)


AnthonyRonning:
@marksftw - this should help the error messages we were seeing. Retries for up to 3 minutes instead of failing after 7 seconds. Users can cancel if that 3 minutes is too long but I errored on the side of longer because agentic runs are starting to get longer and longer, so 3 minutes is less chance a hard stop occurs.

Issue #792, opened 2026-08-18
Agent Mode: optional mid-run steer alongside queue


AnthonyRonning:
## Summary

Maple Agent now has a staged FIFO **queue**: mid-run composer sends stay as chips and are promoted only after the current Goose `Agent::reply` finishes (the whole agentic loop). Some people would rather **steer**: inject follow-ups at the next model turn so the live loop can change course before it finishes every tool.

Queue is a strict improvement even for steer-preferring users. The workaround today is: queue chips, Stop, then Send. That works.

This issue is the follow-up to add steer as a **second** behavior, not to replace queue.

## Why this is not a small flip

Goose already has `Agent::steer` (pinned `064244e`). Maple used it in `14c62af8` and replaced it with the Maple-owned FIFO in `fa85309c` after race/QA issues.

Goose steer:

- does **not** interrupt the current completion or current tool batch
- drains pending steers at the **next** model turn of the **same** `reply`
- persists those messages immediately (`metadata.steer = true`)
- has no public edit-one / remove-one API
- keeps pending steers across cancel for resume

That is earlier than queue, but it is not a chip list. Chips, in-place edit, Discard, and Stop-leaves-chips assume Maple still owns the stack.

## Proposed product shape

Keep **queue as the default**.

Add steer as an explicit second mode (setting TBD: per-task, per-account, or a composer control):

- **Queue:** wait for the current `reply` to finish; chips stay Maple-owned and editable.
- **Steer:** hand the follow-up to Goose at Send; the next model turn sees it; that item leaves the chip list and is no longer in-place-editable.

Do not silently make the current chips steer. ACP should stay out of this unless we explicitly take Goose's unstable `_goose/unstable/session/steer`.

## Out of scope

- Replacing queue
- Interrupting an in-flight token stream or in-flight tool
- ACP steer unless product explicitly wants the unstable Goose wire method

Related: PR #781.

AnthonyRonning:
I think the main question to answer here is: 

1. Do we have both? Codex does. 
2. Which is default?
3. How to surface switching between the two? Settings and/or something in the chat composer view? 
4. Be able to support having both at the same time and doing one-off steer/queue

AnthonyRonning:
Shipped the Maple-owned **queue** in https://github.com/OpenSecretCloud/Maple/pull/781.

[...]

PR #794, opened 2026-08-19
feat(agent): steer a queued follow-up into the current turn

AnthonyRonning:
## Summary

Maple Agent still **queues** on Send. This adds an explicit **steer** path so a follow-up can join the live turn instead of waiting for the whole agentic loop.

- Chip **↑** hands that one chip to Goose `Agent::steer` as soon as the current model + tool batch finishes. Other leftover chips stay queued.
- **Cmd/Ctrl+Enter** with a typed draft steers that draft (no chip).
- **Cmd/Ctrl+Enter** with an empty composer steers the **whole leftover chip stack** in FIFO order. Regular Enter / Send still only queues.
- Idle ↑ on a leftover chip starts a run with **only** that message; remaining chips stay staged.
- Stop still cancels Goose only. Already-steered text stays on the timeline like a normal sent user message (Maple persists it, then discards Goose's in-memory pending steers so they are not re-injected on the next reply).
- Pending `...` stays **between** the original first-send user row and a fast steer. It does not vanish, and it does not move under the steered row.

Queue remains the default. ACP is unchanged and still rejects a second prompt.

Relates to #792.

## Test plan

- [ ] Mid-run Send still creates a chip; leftover chips flush after the current reply finishes.
- [ ] Chip ↑ removes that chip, shows the steered user row, and the current run continues (no second reply).
- [ ] Fast steer after first Send: `...` stays between the first user message and the steered message.
- [ ] No extra `...` under a steered row while the previous assistant is still streaming / steer is waiting for Goose.
- [ ] Cmd/Ctrl+Enter with a typed draft steers that draft without creating a chip.
- [ ] Mid-run: queue one or more chips, leave the composer empty, Cmd/Ctrl+Enter — every leftover chip is steered in order and the chip row clears.
- [ ] Empty Cmd/Ctrl+Enter does nothing when there are no leftover chips.
- [ ] Empty Cmd/Ctrl+Enter does not fire while a chip is being edited.
- [ ] Stop after a steer: steered text remains; leftover chips remain; next Send does not re-inject the steered text as a new user message.
- [ ] Idle leftover chip ↑ starts only that message; other chips stay.
- [ ] Pencil / Discard / X on leftover chips still work after a steer.

### DIFF
--- a/docs/agent-mode-acp.md
+++ b/docs/agent-mode-acp.md
@@ -125,7 +125,7 @@ The refactor exposes a transport-neutral Maple service around the existing embed
 - ensure the account-scoped runtime exists;
 - create and delete a Maple task;
 - send and cancel a run;
-- stage a Desktop follow-up onto Maple's per-task FIFO and promote it with a later sequential `Agent::reply` on the same outer run; Stop cancels only the live Goose turn and leaves staged chips; X drops a chip; pencil edits a chip in place; an open edit holds leftover promotion until save or discard; Discard leaves that chip unchanged; a later Send flushes leftover chips plus any composer draft;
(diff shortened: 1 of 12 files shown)


AnthonyRonning:
@marksftw this is pretty solid if you want to play around with it, I'm curious your thoughts. 

I really enjoy the queue by default, with easy button send button or cmd+enter to steer asap. 

Queue waits until the agent finishes running before sending the message. Useful when you know you will have follow up work after the current instructions complete.

Steer sends user message after the next finished assistant tool call or content.

PR #799, opened 2026-08-19
feat(agent): expand ACP support for Paseo

AnthonyRonning:
## Summary

This draft expands Maple's existing ACP preview into a Paseo-compatible edge over the existing authenticated `MapleAgentService`. It intentionally keeps normal Maple Desktop Agent Mode on the direct embedded-Goose path.

The current snapshot adds:

- dynamic authenticated Maple model discovery and pre-first-message selection;
- ACP session list, load, close, ordered replay, per-turn usage, stable message IDs, and richer tool lifecycle projection;
- caller-owned one-shot permission handling and stronger session/run/lease cancellation fencing;
- provisional ACP task cleanup and probe hygiene;
- a lease-scoped, loopback-only Streamable HTTP MCP client that stays outside Goose's extension registry, OAuth store, and persisted MCP configuration;
- one static ask-before `external_mcp` wrapper with exact frozen routing to injected Paseo tools;
- bounded connection parallelism, a non-GUI `--version` fast path for provider diagnostics, and updated ACP/Paseo documentation.

## Draft status

This is a preservation and review checkpoint, not a merge-ready claim.

- The implementation is based on `c0b3f607` and is currently 26 commits behind `master`.
- It has deliberately **not** been rebased yet, so independent review can inspect the exact locally tested snapshot before conflict resolution.
- A current-master rebase/update, conflict review, complete validation, and renewed Paseo end-to-end smoke are still required.
- Generic caller-supplied stdio MCP remains rejected because accepting it would execute arbitrary local code before a Maple permission boundary. The injected Paseo path is loopback HTTP only.
- Images, audio, embedded resources, SSE MCP, resume/fork/delete, and Windows ACP UI exposure remain outside this draft's supported surface.

## Security and architecture notes

Transient Paseo MCP definitions, authorization headers, and clients are held only by the revocable calling-surface tool context. They are not registered with Goose's extension manager, do not use Goose OAuth fallback, and are not serialized into Maple session configuration. Goose sees only the static `external_mcp` wrapper, which remains in the explicit ask-before policy.

Review attention is especially welcome on:

- lease publication, cancellation, disconnect, close, and dropped-future cleanup;
- preservation of Desktop task ownership and permission mode;
- provisional-task deletion conditions and load/close concurrency;
- transient MCP credential lifetime, result/catalog bounds, and exact routing;
- ACP history/tool projection and compatibility with Paseo's generic adapter.

## Validation at this checkpoint

The repository pre-commit gate passed on macOS/aarch64:

- Prettier format check;
- TypeScript build and Vite production build;
- 661 Bun tests;
- Rust `cargo test --all-targets`: 334 library tests passed, 2 ignored, plus 3 binary tests passed;
- `git diff --check`.

Development-time Paseo smoke exercised the exact workspace Maple executable, dynamic provider/model discovery, normal chat, caller-owned permissions, and injected Paseo MCP tool routing. That runtime evidence must be repeated after updating onto current `master`; it is not packaged or cross-platform release evidence.

### DIFF
--- a/docs/agent-mode-acp.md
+++ b/docs/agent-mode-acp.md
@@ -76,7 +76,7 @@ Use the active authenticated Paseo daemon for end-to-end testing:
 1. Launch the exact Maple build that Paseo will invoke and sign in.
 2. Ensure Maple Agent Mode is running.
 3. Enable the `agent_connections` feature and open **Settings -> Agent connections** in Maple.
-4. Leave **ACP client decides** selected, save the policy if needed, and start the ACP service.
+4. Leave **ACP client decides** selected, save the policy if needed, and start the ACP service once.
 5. In Paseo Desktop, add or edit a Generic ACP provider. Set Command to the exact Maple executable and add `acp` as its one separate argument.
 6. Enable MCP-server support for that provider. Do not add a Maple API key or copy Maple credentials into Paseo.
 7. Let Paseo discover Maple's authenticated model list and interactive mode. A static model list is unnecessary.
@@ -92,7 +92,7 @@ VITE_FORCE_FEATURE_FLAGS=agent_connections
 
 Restart the frontend development process after changing the flag. On macOS, use the executable belonging to the exact development app being tested; do not point Paseo at an installed release while validating a development build.
 
-The settings surface is available only in macOS and Linux Tauri Desktop builds and is disabled by default. Web, mobile, and Windows builds do not expose it. The feature flag controls discovery of the settings page, not service activation. ACP must be started manually after each Maple launch.
+The settings surface is available only in macOS and Linux Tauri Desktop builds and is disabled by default. Web, mobile, and Windows builds do not expose it. The feature flag controls discovery of the settings page, not service activation. After the user starts ACP once, Maple persists that enabled choice and restores the listener after a later authenticated Desktop launch. **Stop service** persists the disabled choice and prevents that restoration.
 
 The default maximum is eight simultaneous ACP connections. The current settings UI exposes neither that limit nor the allowed-project-root list. Its saved default has an empty root list, which accepts any absolute working directory Maple can access. The backend configuration still validates both fields, and changes require **Stop -> Save -> Start**, but the preview UI must not be described as a root-policy editor.
 
--- a/docs/agent-mode-acp.md
+++ b/docs/agent-mode-acp.md
@@ -1,289 +1,333 @@
-# Agent Mode ACP harness preview
+# Maple Agent Mode ACP adapter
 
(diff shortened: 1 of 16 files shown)


AnthonyRonning:
**Security follow-up for later evaluation (non-blocking): macOS `verify_connector_endpoint` is a no-op**

One medium-severity item from a security review of this change. Not proposing we hold this PR on it, but worth capturing while it's fresh so we can evaluate it as a follow-up.

**The gap**

`verify_connector_endpoint` only hardens the connector on Linux (`frontend/src-tauri/src/agent_acp.rs:3550`). On non-Linux targets it's a stub that returns `Ok(())` unconditionally (`agent_acp.rs:3567`), so the `maple acp` connector connects to whatever is listening at the endpoint without checking it.

**Attack scenario (same-user)**

- The endpoint path is derivable: on macOS it's `$TMPDIR/maple-acp-<sha256(exe-path)[:8] hex>.sock` (`endpoint_path()`, `agent_acp.rs:2673`). `$TMPDIR` is per-user, so this is limited to same-user processes — but any same-user process can predict the path.
- A same-user process can pre-bind/squat that path before Maple Desktop's ACP service starts. `bind_listener` fails closed against a live listener (“Another Maple ACP service is already listening”, `agent_acp.rs:2661`), so the squatter wins the race and Maple never binds.
- The connector then sends `_maple/bridge/hello` — carrying the allowlisted env (`ALLOWED_BRIDGE_ENV`, `agent_acp.rs:64`, which includes `BUZZ_PRIVATE_KEY`, `BUZZ_AUTH_TAG`, `BUZZ_API_TOKEN`, `BUZZ_RELAY_URL`) — followed by all subsequent ACP prompt traffic, to the squatter.

**Why it's worth revisiting now**

This is pre-existing (introduced with the exploratory harness in 95895dfd), not a regression from this PR. But this change raises the stakes of the endpoint: it now gates session listing/loading, task history, and up to 8 concurrent connections, whereas the harness exposed much less.

**Honest caveats**

- Requires attacker code already running as the same user — the weakest trust boundary, and much of what it exposes (session history) is already on a same-user-readable disk.
- The real delta is the bridge-hello credentials and live traffic interception (same-user processes can't trivially read another process's env on macOS), plus the ability to drive sessions while the service is running.
- ACP service is manual-start only (no auto-start), which narrows exposure.

**Suggested fix (small)**

Drop the `#[cfg(target_os = "linux")]` gate and run the same checks on macOS: `MetadataExt::uid()` and `permissions().mode() & 0o077` work there, and the existing Linux-side test can be mirrored. Roughly 15 lines.

Flagging for evaluation — fine with landing this PR without it.

PR #813, opened 2026-08-21
Ask before installing updates on Linux deb/rpm

benthecarman:
On Linux, deb and rpm updates install through pkexec, which opens a system password dialog. The automatic updater ran that install with no warning, so the dialog appeared with no context for the user.

Download the update as before, but on deb/rpm builds hold it in memory and show an "Update Available" toast instead of installing. The install, and with it the password prompt, runs only when the user clicks Install Now. The existing "Update Installed" and "Update Failed" toasts follow from there, and the hourly check does not download the same version again while it waits for approval.

AppImage, macOS, and Windows still install silently because they show no prompt.

Preview: 
image

### DIFF
--- a/frontend/src-tauri/src/lib.rs
+++ b/frontend/src-tauri/src/lib.rs
@@ -100,6 +100,62 @@ fn get_pending_update_failure() -> Result<Option<String>, String> {
 }
 
 #[cfg(desktop)]
+#[tauri::command]
+fn get_pending_update_install() -> Result<Option<String>, String> {
+    Ok(PENDING_UPDATE
+        .lock()
+        .map_err(|e| format!("Failed to lock PENDING_UPDATE mutex: {e}"))?
+        .as_ref()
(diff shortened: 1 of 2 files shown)


AnthonyRonning:
Thanks for addressing the unexpected password prompt. One separate packaging detail surfaced while reviewing this PR: Maple’s current RPM release is rebuilt from the Debian payload, and the v3.3.5 RPM binary contains Tauri’s Debian bundle marker. Because Tauri uses that marker to choose both the update artifact and installer, an RPM installation appears to select the `.deb` update and attempt `dpkg` rather than the RPM path.

Did you encounter the prompt on an RPM-based installation, and have you tested this change using Maple installed from the actual `.rpm` artifact? I believe the marker problem predates this PR—the consent change itself still looks useful—but I want to distinguish “defer the privilege prompt” from “the RPM updater works end to end.”

AnthonyRonning (approved):

PR #950, opened 2026-09-16
Make Tab reach the next login field in maple-gpui

benthecarman:
## Summary

Pressing Tab in the maple-gpui sign-in email field did nothing, so the password field could only be reached with the mouse.

gpui applies an element's `tab_index` only to a focus handle it creates itself. `TextInput` tracks its own explicit handle, so the div-level tab index was ignored, every input stayed a non-tab-stop, and `window.focus_next` found nothing to move to.

The fix puts the tab index and tab-stop flag on the input's focus handle in `with_tab_index`. This also restores tab order for the other inputs built with the same helper (settings, sidebar, API keys).

Enter in the password field already submitted; the new test confirms it.

## Test

Added `tab_reaches_password_and_enter_submits` in `ui/login.rs`. It mounts the real login screen with the shipped keymap, types an email, presses Tab and checks the password field has focus, Shift+Tabs back to the email field, then types a password and presses Enter and checks the screen entered the signing-in state. It failed on the Tab step before the fix.

`cargo fmt --check`, `cargo clippy -p maple-gpui --all-targets -- -D warnings`, and `cargo test -p maple-gpui` (309 tests) pass.


AnthonyRonning (approved):

AnthonyRonning (changes requested):
Failing on linux: 

```
test ui::settings::navigation::tests::picking_from_the_dropdown_persists_to_disk ... ok
test ui::spell::tests::suggests_replacements ... ok

failures:

---- ui::login::tests::tab_reaches_password_and_enter_submits stdout ----

thread 'ui::login::tests::tab_reaches_password_and_enter_submits' (15408) panicked at app/src/ui/login.rs:454:13:
enter in the password field must submit


failures:
    ui::login::tests::tab_reaches_password_and_enter_submits

test result: FAILED. 314 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.78s

error: test failed, to rerun pass `-p maple-gpui --bin maple-gpui`
error: recipe `test` failed on line 40 with exit code 101
```

benthecarman:
Fixed the flaky Enter assertion: `simulate_keystrokes` runs until parked, so the unreachable test backend could answer before the check. The test now asserts the validation error on an empty password, then accepts either in-flight or failed for the real attempt. Squashed into the single commit.

AnthonyRonning (approved):

Issue #943, opened 2026-09-15
External agent tools never appear in sessions that existed before the Codex integration was enabled


AnthonyRonning:
## Summary

The external-agents gating comment in `crates/maple-agent/src/agent.rs` (~line 8600, the `.with_external_agents(...)` filter) says:

> The toggle is read at every agent build, so it takes effect at the next run.

Observed behavior contradicts this: a session that existed before the Codex integration was enabled never gains the five external-agent tools (`list_agent_providers`, `agent_start`, `agent_send`, `agent_status`, `agent_cancel`), even after the app is fully rebuilt, relaunched, and the session is resumed and continued with new messages.

## Repro

1. Have a chat session created before the Codex integration existed / was enabled.
2. Enable Codex under Settings > Integrations.
3. Relaunch the app, resume the old session, send a message.
4. The task's tool list still lacks `agent_start` etc.

## Expected

Per the code comment: the toggle is re-read at every agent build, so a resumed pre-existing session should gain the tools on its next run after the integration is enabled.

## Actual

Only sessions created *after* the integration was enabled get the tools. A new session works correctly (Codex detected, signed in, tools usable).

## Suspected cause

The developer extension's tool list is likely snapshotted into Goose's persisted session/extension state at session creation (sessions.db) and not refreshed on later agent builds for that session — similar to other per-session frozen config. The `external_agents_enabled(...)` check does run at agent build time, but its result may never reach a session whose developer extension was already materialized.

## Workaround

Start a new task after enabling the integration.

## Verified environment

- Binary at commit 4178b1cc
- `integrations.json` shows `"id": "codex", "enabled": true, "backend": "embedded"`
- Codex 0.154.0 signed in; new sessions work end to end

benthecarman:
Couldn't reproduce on current master (`0b0736ba`): a task created with Codex disabled gained the external-agent tools after enabling Codex and restarting the app.

PR #954, opened 2026-09-16
Add per-task integration selection

benthecarman:
The composer now shows Codex and CUA, including a Settings link when setup is needed. External-agent selections persist per task; tasks without an explicit choice inherit current Settings defaults on each run, including after restart.

Provider metadata is shared across Settings, composer rows, and tool admission. Future providers can reuse the selector and persistence by adding discovery and a transport adapter. Typed selection kinds keep provider IDs separate from custom MCP names, and runtime admission is restricted to the task's selected providers.

### Validation

- Component `just ci` passed: formatting, feature-matrix clippy, warning-denied builds, and 801 passing tests (2 ignored).
- Native Linux release compiled without Nix. Computer-use checks passed for selector visibility, real provider-list calls, disabling/re-enabling, restart persistence, inherited defaults, explicit overrides, and CUA setup navigation.
- Runtime regressions cover warm/cold tool catalogs, persisted old sessions, ACP exclusion, and rejection of unselected providers.

### Remaining limitations

Related to #943. The original missing-tools behavior did not reproduce on clean master `0b0736ba`: a task created with Codex disabled gained the external-agent tools after enabling Codex and restarting. This PR establishes selector behavior and regression coverage, without claiming a confirmed root-cause fix for that report.

Full Codex delegation remains unverified: both master and this branch launched Codex but hung during external-agent startup. CUA execution was unavailable because the test desktop requires its GNOME helper; the setup link was verified.

### Screenshot

Codex and CUA in the task integrations selector


AnthonyRonning (changes requested):
cicd

AnthonyRonning (approved):

PR #958, opened 2026-09-17
Fix initial sidebar scroll position

benthecarman:
When tasks load after the empty sidebar has rendered, GPUI's first splice moves the scroll anchor past the inserted rows and opens the sidebar near the bottom.

Initialize the first list contents at the top, regardless of load/render order. Keep incremental splices for subsequent updates so they preserve the user's scroll position.

Validation: Agent `just ci`, formatting, and desktop tests passed. Checked delayed loading and wheel scrolling in a native GPUI sidebar fixture with 200 synthetic tasks; authenticated startup was not exercised.


benthecarman:
rebased on master to fix ci

AnthonyRonning (approved):

PR #973, opened 2026-09-18
OpenSecret: standard Flash circuits and safe inference diagnostics

### LINKED ISSUE #970: OpenSecret: log bounded, sanitized upstream inference error details
Upstream inference failures currently retain status and selected safe headers but discard the response body. An HTTP 400 therefore cannot be distinguished as a context-limit error, unsupported parameter, malformed tool history, or another provider rejection using the existing execution diagnostics.

Add a short, sanitized upstream error message summary or snippet, plus an error code/type when available, correlated with the existing request, execution, attempt, provider, model, and HTTP-status metadata.

### Scope and acceptance criteria

- Read only a bounded prefix of non-success inference responses, with explicit byte and time limits. Handle both attested and standard provider transports through the existing provider boundary.
- Prefer allowlisted structured error fields. Sanitize and truncate the message; never log raw bodies, echoed prompts/responses, tool arguments, credentials, headers, or other customer content. Use a safe category/code-only fallback when a message cannot be safely summarized. Strip control characters to prevent log injection.
- Record whether diagnostics were truncated, unavailable, or unparseable. Empty, oversized, non-JSON, invalid-encoding, stalled, and interrupted bodies must still produce the original provider failure promptly.
- Diagnostic collection must preserve cancellation, one terminal outcome, status/Retry-After classification, circuit signals, usage accounting, and existing client-visible error contracts. Do not introduce inference replay or provider retries.
- Add focused tests for useful sanitized diagnostics and the privacy/resource boundaries above, including provider error messages that echo sensitive input.
- Update relevant logging guidance narrowly to distinguish safe extracted diagnostics from prohibited raw provider-body logging.

Starting points: `try_provider`, `attempt_failure_from_provider_error`, and `log_attempt_terminal` in `services/opensecret/src/web/openai.rs`; `UpstreamProviderError` and the response abstraction in `services/opensecret/src/provider_client.rs`.

The current discard behavior is explicit in [the provider error path](https://github.com/MaplePrivacyLabs/Maple/blob/1ef3daaacd00d38989c4ac08bd933c877edee909/services/opensecret/src/web/openai.rs#L3905-L3924). This issue improves diagnostics; it does not presume that an upstream 400 is a routing defect.

### LINKED ISSUE #971: Router V2: apply standard circuit breakers to GLM 5.3 Flash
Router V2 assigns GLM 5.3 Flash `FailoverPolicy::CapacityGates`, while ordinary completion models use `AllGates`. Flash therefore excludes providers for 429/503/529 capacity signals but ignores route-health circuits triggered by repeated timeouts, transport failures, or stream failures. A provider can remain eligible for Flash even when its route-health circuit is open.

Use the normal `AllGates` behavior for GLM 5.3 Flash. Reliability should take priority over avoiding the cost of an alternate provider during failures. The healthy allocation and the failure policy are separate decisions.

### Scope and acceptance criteria

- Register Flash with the ordinary completion-model failover policy. Remove Flash-specific capacity-only machinery if it has no remaining consumers.
- Apply the standard route-health gate consistently during both initial provider selection and the send-time claim. Cover both Tinfoil and Continuum directions.
- Reuse existing failure thresholds, success reset, cooldowns, and fenced recovery probes. A single ordinary timeout still follows the standard threshold; capacity signals retain their existing behavior.
- Preserve the existing healthy provider weights, sticky account allocation, public model identity, pricing/usage attribution, V1 routing, and feature-flag boundaries. Do not add a model-selector flag.
- Affect subsequent eligible requests only. Do not replay the request that timed out or failed, including after partial output or ambiguous acceptance.
- Replace tests that require Flash to ignore route-health gates with focused tests for ordinary timeout/transport/stream circuit behavior, alternate-provider selection, recovery, and both-routes-unavailable behavior. Preserve coverage for neutral HTTP 400/client-cancellation outcomes and capacity gates.
- Ensure Auto routing and any related in-flight implementation/tests consume the revised shared policy instead of retaining a Flash-specific exception.

Starting points: `services/opensecret/src/provider_registry.rs`, `src/provider_routing.rs`, and `src/inference/health.rs` under that backend component. [The current registry](https://github.com/MaplePrivacyLabs/Maple/blob/1ef3daaacd00d38989c4ac08bd933c877edee909/services/opensecret/src/provider_registry.rs#L237-L245) shows the Flash-only exception; `glm_flash_probe_skips_the_route_health_gate` is one test that encodes it.

Changing global timeout durations, circuit thresholds, healthy weights, or Auto candidate ordering is outside this issue.

AnthonyRonning:
GLM 5.3 Flash now follows the standard Router V2 route-health and capacity gates, so repeated timeouts, transport failures and stream failures can steer later requests to another eligible provider. Auto model selection uses the same availability policy. Healthy weights, V1 routing and the no-replay boundary stay unchanged.

Upstream HTTP errors now carry bounded diagnostics into the correlated attempt warning: an 8 KiB / 100 ms reader emits only recognized codes, types and fixed summaries. Raw provider messages and echoed customer content never enter the diagnostic record. Cancellation during that read preserves the already-known status, retry hint and probe outcome. V2 alternate Auto and sticky/fallback provider decisions also emit correlated INFO records for the route that won the send-time claim.

Validation: format check, Clippy with warnings denied, and the locked all-features test suite (701 passed, 23 intentionally ignored). Added coverage for both provider directions, Auto availability, failure thresholds and recovery, diagnostic privacy/resource bounds, and cancellation after error headers. Independent source review completed. Live providers, authenticated deployed flows and EIF/release validation remain for the release-testing stage.

Fixes #970
Fixes #971


AnthonyRonning:
Reviewed `41fbc3ce` against `d5f359e7` (master including #968) from an isolated detached checkout.

**Verdict: approve. No P0, P1, or P2.** Both issues' acceptance criteria are met, and the subtle part of the change, cancellation during the diagnostic read, is handled and tested.

**Flash circuits (#971)**
- `FailoverPolicy` is removed rather than special-cased: `available_providers_for_model` uses the effective snapshot for every model, and `try_claim_probe_locked` always includes the route-health gate, so selection and the send-time claim agree. Auto model availability consumes the same function, so the Auto path needs no separate change.
- Replacement tests cover both provider directions, all four route-health failure kinds, the three-failure threshold, half-open recovery through a single fenced probe, a failed probe reopening the circuit, neutral 400 and consumer-drop outcomes, success resetting the streak, and V1 independence. The capacity-gate and V1 Flash tests the issue asked to preserve are still present. Healthy weights and the 30% Continuum split are unchanged; the routing policy version bump is appropriate.

**Diagnostics (#970)**
- The reader is a genuinely lossy boundary: only static allowlisted labels and fixed summaries can leave it, message text is matched by anchored prefix and never copied, unknown values are suppressed, and the `Debug` output therefore cannot carry control characters or echoed content. Read is bounded to 8 KiB and 100 ms total with a yield against always-ready streams; reaching the byte bound is treated as truncated without parsing the prefix; dropping the future drops the owned stream. Both transport variants share one path.
- Status, `Retry-After`, capacity classification, the public error shape, and route-health signals are unchanged; the diagnostic only travels into the correlated attempt warning. No retry or replay is introduced.
- Cancellation after error headers now records the already-known upstream failure instead of a neutral consumer-drop, and a held probe lease resolves with it. The test drops the future mid-read for 429/503/529 with and without a recovery lease and asserts exactly one terminal, the retained status and hint, and a still-fenced route.

**Routing INFO record**
- Fires only for non-primary Auto reasons and sticky or fallback routes, after the claim, keyed by request, execution, and attempt ids, with no account identifier. This closes the observability gap noted on #968.

**Notes, none blocking**
- The observation guard's drop still logs "Inference attempt abandoned before terminal processing" when it has just recorded a known provider failure; the wording now overstates what happened.
- The decision-time line that listed the rejected candidates and their retry hints moved to debug, so INFO shows that an alternate ran but not why the preferred model lost. Consider carrying a compact rejected summary on the new INFO record.
- In a Responses tool loop the INFO record repeats once per model turn for a sticky or fallback route. Bounded and joinable, just worth knowing for volume.
- Every failing provider response now waits up to 100 ms for diagnostics before the error is returned and the circuit observes it. Bounded and documented; noted only as a deliberate latency trade.

**Validation executed at `41fbc3ce`**, pinned Nix shell with stateful hooks disabled: `cargo fmt --all -- --check` clean; `RUSTFLAGS='-D warnings' cargo clippy --locked --all-targets --all-features -- -D warnings` clean; `RUSTFLAGS='-D warnings' cargo test --locked --all-features`: 701 passed, 0 failed, 23 ignored (the pre-existing database and live-provider suites). Checkout stayed clean. At my last read, the Nix checks and application package job had passed and the backend format/lint/test and SDK integration jobs were still running.

**Not exercised:** live provider error bodies, deployed flows, and EIF/PCR evidence, consistent with the PR's own scope statement.

PR #976, opened 2026-09-19
Add Claude Code delegation to Maple Agent

benthecarman:
Adds Claude Code delegation alongside Codex in Maple Agent. Tasks can stream activity, answer approval and question cards, resume delegated sessions, and stop their child processes. The native Rust transport is adapted from the pinned Goose provider and requires only the installed Claude CLI.

The Claude card uses its logo and checks `claude auth status --json` to show signed-in, signed-out, or unknown status. Status probes are bounded and account details are not retained.

For both providers, Settings now controls availability in the composer. Each task must explicitly select a provider; disabling it in Settings hides it and blocks saved selections on the next run. Task choices persist across restarts.

Validation:

- Current non-Nix validation: formatting, four Clippy configurations, workspace builds, and all 806 workspace tests passed. The final headless test build and its retry ran out of disk space; that check is incomplete. The earlier revision passed all 839 tests in `just ci`.
- Native CLI fixtures cover streaming, resumption, provider isolation, permissions, cancellation, malformed output, process cleanup, and authentication probe failures/timeouts.
- Settings gates and session tool refresh tested with warm and cold runtimes and ACP exclusion.
- Latest Linux debug app: verified the logo and live sign-in status in light/dark themes and toggled Claude off/on. Authenticated composer interaction was not exercised in the isolated offline UI profile.
- An earlier non-Nix Linux release build passed, and the user tested the uploaded delegation build.

Native GUI checks ran on Linux only.


apps/maple-agent/crates/maple-agent/src/agent/external_agents/mod.rs
```diff
@@ -970,59 +1036,95 @@ impl ExternalAgent {
                 return Ok(());
             }
         }
-        let executable = codex::find_executable(call.login_path.as_deref()).ok_or_else(|| {
-            "Codex is not installed, or `codex` is not on PATH. Ask the user to install the Codex CLI.".to_string()
-        })?;
+        let existing_thread = self.state.lock().await.thread_id.clone();
+        let claude_thread = existing_thread
+            .clone()
+            .unwrap_or_else(claude::new_session_id);
```
AnthonyRonning, line 1038: **A failed first Claude launch strands the agent.** For Claude, `ThreadStart` echoes this locally generated UUID, so `state.thread_id` is committed (below, after the handshake) before the CLI has persisted anything. If the first turn dies at startup (not signed in, an `--effort`/`--model` value the CLI rejects, any usage error), every later `agent_send` on that agent runs `--resume <uuid>`. On 2.1.258 that returns `result subtype=error_during_execution` with stderr `No conversation found with session ID: ...`, which Maple renders as the same generic "could not complete this request" text, so the user cannot tell the agent is permanently broken. Codex only records a thread after `thread/start` succeeds.

Suggestion: commit the Claude session ID only once the child has emitted `system/init` (or the first successful `result`), or fall back to `--session-id` with the same UUID when a resume fails immediately with no turns.

apps/maple-agent/crates/maple-agent/src/agent/external_agents/claude.rs
```diff
@@ -0,0 +1,647 @@
...
+            let questions: Vec<_> = input
+                .get("questions")
+                .and_then(Value::as_array)
+                .into_iter()
+                .flatten()
+                .enumerate()
+                .filter(|(_, question)| question.is_object())
+                .map(|(i, question)| {
+                    let mut question = question.clone();
+                    question["id"] = format!("q{i}").into();
```
AnthonyRonning, line 440: **Multi-select questions become single-select.** `AskUserQuestion` sends `multiSelect: true` per question (verified on 2.1.258), but `parse_user_questions` in `developer_tools.rs` only keeps `id`/`header`/`question`/`options`, so the question card lets the user pick one option. The answer path in `respond` already joins several values with `, `, so only the card side is missing. Not blocking, but worth carrying `multiSelect` through the projection here so the card can allow multiple picks.

AnthonyRonning (commented):
Two non-blocking findings from a local review against claude 2.1.258 (stream-json protocol, interrupt, resume, and AskUserQuestion all verified outside the GUI). Details inline.

apps/maple-agent/crates/maple-agent/src/agent/external_agents/tests/claude_fixture.rs
```diff
@@ -0,0 +1,261 @@
...
+                finish(
+                    &mut out,
+                    session,
+                    if allowed { "Allowed" } else { "Denied" },
+                );
+            }
+            _ => panic!("unexpected Claude fixture input"),
+        }
+    }
+    writeln!(log, "{}", json!({"event": "stdin_closed"})).unwrap();
```
AnthonyRonning: **Fixture log write race breaks `claude_native_retries_only_resume_confirmed_sessions`** (fails on macOS and Ubuntu CI at `tests.rs:1167`, `args.len()` is 2 not 3).

In the `error` iteration Maple kills the fixture as soon as it reports the failed result, which is the intended runtime behaviour. The fixture is mid-way through this `stdin_closed` write when the kill lands. `writeln!(log, "{}", json!(..))` on an unbuffered `File` goes through serde's `Display`, which emits the value in several small `write` calls, so the kill leaves a partial `{"event"` with no trailing newline. The next process then appends its `args` record to that same line. From the log on my machine:

```
{"event"{"args":["--input-format","stream-json",...,"--resume","d572c147-..."],"cwd":".../project/sub"}
```

That glued line is not valid JSON, so the test's `filter_map(serde_json::from_str)` drops the second launch record. The retry logic itself did the right thing (one `--session-id` launch, then two `--resume` launches on the same UUID).

Fix in the fixture: format each log line to a `String` first and write it with one call, here and for the `args` record at the top of the loop (line 116 has the same latent risk):

```rust
let line = json!({"event": "stdin_closed"}).to_string();
writeln!(log, "{line}").unwrap();
```

A single `write` of one line is atomic enough for this purpose on the platforms the test runs on. Alternatively, log `stdin_closed` before leaving the read loop rather than after, but the single-write form also protects the other records.

AnthonyRonning (changes requested):
The two fixups resolve both earlier comments (session confirmation before `--resume`, and `multiSelect` through to the card), thanks. Requesting changes only because the new `claude_native_retries_only_resume_confirmed_sessions` test fails deterministically on macOS and Ubuntu CI (Windows passes because the fixture tests are Unix-only), and I reproduced it locally on macOS 4/4 runs. It is a fixture log race, not a runtime bug; details inline.

apps/maple-agent/crates/maple-agent/src/agent/external_agents/tests/claude_fixture.rs
benthecarman: fixed

AnthonyRonning (approved):
Fixture fix confirmed: `claude_native_retries_only_resume_confirmed_sessions` and the other six `claude_native_*`/detection tests pass locally on macOS at 7a9b23c7 (previously failed 4/4 runs). Single-write log records are the right fix and the stricter parser will catch any regression. Earlier comments on session confirmation and multiSelect are resolved. Approving; CI on this head still needs a workflow approval click.

AnthonyRonning (approved):

